PR #72, opened 2026-03-24
Rewrite coding agent prompt for faster, higher-quality demos

strickvl:
## Summary

- Rewrites the coding agent system prompt to produce impressive demo results faster
- Adds an explicit **5-8 tool-call budget** so the agent finishes efficiently instead of filling the 30-round limit
- Restores the **parallel tool-call instruction** (removed in 30eb769) — lets the dashboard show multiple checkpoints appearing simultaneously while using fewer LLM rounds
- Adds **visualization rules** (clean backgrounds, curated palettes, readable fonts) so Plotly charts look publication-quality on the first attempt
- Removes "think step by step" and verbose guidelines that encouraged unnecessary intermediate rounds
- Restructures into scannable TOOLS / VISUALIZATION / GENERAL sections

## Context

During demos today we noticed the agent was too slow to reach a result — a combination of too many clarifying questions, too many tool-call rounds, and underwhelming first output that needed follow-up iterations. This prompt rewrite targets all three issues.

## Test plan

- [ ] Run `cd examples/coding_agent && python agent.py "Create a Plotly population pyramid for South Korea"` and verify it completes in ~5-8 checkpoints
- [ ] Verify the generated chart has clean styling (white background, labeled axes, curated colors)
- [ ] Verify the agent calls `hand_back` at the end rather than just responding with text
- [ ] Verify parallel tool calls still produce concurrent checkpoints in the dashboard


strickvl:
Low quality PR.

PR #119, opened 2026-04-09
Add cross-session memory, compaction, and memory tools to coding agent

AlexejPenner:
## Summary

Adds persistent cross-session memory and conversation compaction to the coding agent example, demonstrating `kitaru.memory` primitives in a real agent loop.

- **New `agent_memory.py` module** — session context loading (past task summaries injected at flow start), LLM-facing memory tools (`remember`/`recall`/`list_memories`) executed in the flow body, and conversation compaction via `KitaruClient.memories.compact()` when messages exceed a configurable threshold
- **Agent loop refactored** — tool dispatch split into checkpoint tools (parallel, replayable) and flow-body tools (memory ops + HITL), with extracted helpers for message building, tool call parsing, and each dispatch path
- **Memory tool schemas and prompts** — `RememberParams`/`RecallParams`/`ListMemoriesParams` added to `tools.py`; system prompt updated with memory guidelines and key namespace conventions (`prefs/`, `facts/`, `conventions/`, `tasks/`)
- **Transient retry logic moved to `tools.py`** — `dispatch_tool_with_retries()` and `_is_transient_error()` relocated from `agent.py` for better separation of concerns

### Design notes

Memory tools run in the flow body (not inside `@checkpoint`) because `memory.set()`/`memory.get()` are forbidden inside checkpoints — memory ops aren't idempotent and would break replay semantics. The `_resolve_recall` and `_resolve_list` helper checkpoints exist solely to materialize the lazy `OutputArtifact` references that `memory.get()`/`memory.list()` return inside a `@flow`.

Conversation compaction uses `KitaruClient.memories.compact()` to LLM-summarize older messages when the list grows past `COMPACT_THRESHOLD` (default 20), keeping the system message and the last 6 messages intact.

## Test plan

- [ ] Run `python agent.py "Create a simple chart"` and verify memory tools appear in the tool list
- [ ] Verify `remember`/`recall` round-trip: agent stores a preference, subsequent recall retrieves it
- [ ] Run a multi-task session (hand_back → follow-up) and confirm task summaries are saved to memory
- [ ] Trigger compaction by running enough turns to exceed the threshold — check logs for `compaction_status=compacted`
- [ ] Replay an execution and verify checkpoint tools replay correctly while memory state persists


examples/coding_agent/agent.py
```diff
@@ -361,6 +411,14 @@ def _make_display_name(tool_name: str, llm_name: Any, counter: int) -> str:
 @click.command(help="General-purpose interactive agent.")
 @click.argument("task")
 def main(task: str) -> None:
+    kitaru.configure(
+        stack="local_remote",
```
strickvl: this is annoying btw :)

strickvl: add a TODO to remove this before shipping

strickvl:
No longer relevant following #289

PR #226, opened 2026-04-24
Stream SDK logs while waiting for remote flows

### LINKED ISSUE #202: Remote `.run()` submissions stop emitting logs after `Code download finished` until the run completes
## What's happening

When running a flow against a remote (e.g. Kubernetes) stack, the local Python process that called `.run()` prints lifecycle logs up to `Code download finished`, then stops producing terminal output. The run continues on the cluster — checkpoints are created, the dashboard shows live progress — but nothing streams back to the submitter's terminal until the whole run completes, at which point the logs arrive in one flush.

```
Kitaru › Waiting for orchestrator job to finish...
Kitaru › Downloading code from artifact store path s3://…
Kitaru › Code download finished.
                                                   ← terminal sits here
                                                     for the whole run
```

Meanwhile the dashboard shows the orchestrator pod running and emitting new checkpoints in near-real-time.

## Repro

1. `kitaru stack use my-k8s-stack`
2. Run any long-running flow (e.g. `python examples/news_scout/scout.py`)
3. Observe that the submitter terminal stops producing new output after `Code download finished`
4. Open the dashboard URL that was logged — see checkpoints being created live
5. Wait for the run to finish. All the intermediate logs arrive in the terminal at once.

## Expected

The `.run()` submitter should stream orchestrator / step pod output live — or at least surface progress markers (checkpoint started/finished events) — so users know the run hasn't hung.

## Why it matters

Users who don't already know about `kitaru executions logs --follow <exec_id>` will reasonably assume the run has hung, kill the terminal, and lose their visibility. It's also a surprising UX delta from the local-stack path, where `.run()` shows Kitaru-branded lifecycle logs live.

## Possible directions

The submitter already knows the execution id the moment `.run()` accepts it, so it has everything `kitaru executions logs --follow` already uses:

1. **Auto-invoke the tail-from-backend loop inside `.run()` for remote stacks** — reuse existing plumbing, no new capture path needed.
2. **Emit progress markers** (checkpoint started/finished) to stdout from whichever runtime component is polling the orchestrator job, rather than buffering and flushing at the end.

## Environment

- Kitaru: `develop` (0.5.1 + unreleased changes)
- Stack: Kubernetes (deployment type `kubernetes`)
- Kitaru server: staging.cloudinfra.zenml.io
- Python: 3.12

strickvl:
## What changed

- Extracted the CLI follow-mode polling/deduplication loop into a shared client helper.
- Updated `FlowHandle.wait()` to follow live backend step logs while waiting for remote executions, with controls:
  - `stream_logs=False` restores the previous fully silent polling behavior.
  - `log_source`, `log_checkpoint`, and `log_interval` tune backend log following.
  - `log_sink` lets SDK callers explicitly capture raw `LogEntry` objects.
- Made SDK output library-safe by default:
  - `wait()` auto-prints followed log messages only when stdout is interactive.
  - non-interactive/no-sink contexts now use normal poll-only waiting, avoiding hidden backend log-retrieval traffic.
  - non-interactive callers can still opt into live log streaming by passing `log_sink=...`.
- Kept `kitaru executions logs --follow` on the same shared follow loop so CLI and SDK behavior stay aligned.
- Added regression coverage for interactive/non-interactive output, explicit sinks, deduplication, opt-out behavior, unavailable-log fallback, client-construction fallback, invalid wait-log arguments, streaming failure handling, invalid intervals, and sink error propagation.
- Documented SDK wait log following, including the remaining pre-handle submission boundary, and updated the changelog.

Fixes #202.

## Why

Issue #202 showed that a remote Kubernetes flow could be visibly running and producing checkpoint logs, but Python callers using `handle.wait()` saw no progress until the run finished. The live reproduction confirmed that the backend log API and `kitaru executions logs --follow --source step` were already able to stream the checkpoint logs; the missing piece was the SDK wait path, which only status-polled and never tailed logs.

The fix makes the SDK wait path use the same log-follow mechanism as the CLI when there is a real log destination. In an interactive terminal, a normal Python run now behaves like:

```python
handle = my_flow.run()
result = handle.wait()  # follows checkpoint logs and auto-prints in interactive terminals
```

For non-interactive/library contexts, the default remains quiet and poll-only. Callers can use `log_sink=...` to explicitly receive log entries without corrupting structured stdout:

```python
logs = []
result = handle.wait(log_sink=logs.append)
```

## Key implementation details

- The shared follow loop lives in `src/kitaru/_client/_follow.py`.
- CLI formatting remains CLI-owned through a small sink wrapper in `src/kitaru/_cli/_executions.py`.
- SDK wait formatting remains SDK-owned through a sink in `src/kitaru/flow.py`.
- `FlowHandle.wait()` only enters log-follow polling when stdout is interactive or the caller supplied `log_sink`; non-interactive/no-sink waits use the old status-polling path.
- `FlowHandle.wait()` validates wait-log arguments before constructing the Kitaru client when streaming is active, so invalid caller inputs still surface as usage errors.
- Runtime-log/client availability failures fall back to old poll-only waiting; user `log_sink` exceptions remain visible and are not swallowed as fallback.
- Log following starts after `.run()` returns a `FlowHandle`; this PR does not add a pre-handle submission monitor.

## Validation

- Live Kubernetes reproduction on `aws-k8s-stack` before the fix:
  - `FlowHandle.run()` returned quickly.
  - `handle.wait()` stayed silent.
  - `kitaru executions logs --follow --source step` streamed checkpoint logs for the same execution.
- `uv run pytest tests/test_flow.py tests/test_cli.py tests/test_client.py` — 499 passed.
- `uv run pytest tests/test_flow.py -k "flow_handle_wait"` — 18 passed after the final polling guard.
- `just check` — passed after the final polling guard.
- `just test` — passed after the final polling guard: 1584 passed, 8 skipped.

## Reviewer Notes

- Please sanity-check the SDK default matrix:
  - interactive stdout + no sink: follow logs and print messages.
  - non-interactive stdout + no sink: do not follow logs; use normal status polling.
  - explicit `log_sink`: follow logs and deliver `LogEntry` objects, even in non-interactive contexts.
- The follow loop still polls the current log endpoint and dedupes by log-entry content because there is no stable backend log cursor exposed yet. That matches current CLI behavior; a cursor/event-id would be a good future backend/API improvement.
- Pre-handle submission silence is documented but not solved here.


strickvl:
Closing this PR after the follow-up investigation on #202.

The implementation here adds SDK/backend log-following while waiting, but it does not actually solve the behavior that prompted #202 for the AWS Kubernetes stack. In that stack, ZenML submits the Kubernetes orchestrator job asynchronously from the local process and prints the `kubectl logs job/...` hint. The local process is not receiving live step pod lifecycle logs in the same way as local/local-remote execution.

In testing, the PR branch entered the wait/log-follow path, but it surfaced ZenML response hydration warnings such as status/end-time changes rather than checkpoint lifecycle messages. That means the missing checkpoint progress is not just a display-formatting issue; the relevant live checkpoint events are not available through this wait path for the async Kubernetes mode.

Given that this appears to be normal ZenML Kubernetes behavior rather than a Kitaru bug, we are closing #202 as expected behavior and closing this PR as no longer needed. A future product feature could add an explicit Kitaru progress monitor based on step-run state transitions, but that would be a separate design decision rather than this bug fix.

Issue #215, opened 2026-04-23
Service Accounts + API Keys in the SDK


AlexejPenner:
**What problem does this solve?**

**Describe the solution you'd like**

PR #230, opened 2026-04-24
Add auth management surface for service accounts and API keys

### LINKED ISSUE #195: Introduce sdk surface for service accounts/api keys/ PAT

strickvl:
## Summary

Adds an OSS-first auth management surface for Kitaru server credentials, so users can create and manage service accounts and API keys from Python and the CLI without dropping down to the underlying ZenML client.

- **SDK:** new `KitaruClient.auth` namespace with `service_accounts` and `api_keys` sub-namespaces (create / list / show / update / delete / rotate / deactivate / reactivate). Public DTOs `AuthServiceAccount`, `AuthAPIKey`, `AuthAPIKeyWithValue` are re-exported from `kitaru`.
- **CLI:** new `kitaru auth service-accounts …` and `kitaru auth api-keys …` subcommand trees, honouring the shared `--output json` contract (`{command, item}` / `{command, items, count}`).
- **Docs:** new `docs/content/docs/guides/authentication.mdx` walks through the credential chain (service account → API key → `kitaru login --api-key` → short-lived bearer via `kitaru auth token`), plus updates to deployment / Docker / Helm / configuration docs so readers can wire the right env vars.
- **Smoke test:** extended `scripts/smoke-test.sh` to exercise the new CLI surface non-destructively.
- **Analytics:** new `AnalyticsEvent` entries for each auth lifecycle point, wired into both CLI and SDK surfaces.

### Why it matters

Closes #195 (and the related duplicate #215). Previously there was no supported way for an OSS Kitaru user to provision a machine credential for CI or a container without reaching into ZenML-internal APIs. This gives them one.

### Key design decision: the raw key is only returned once

List / show / update responses are metadata-only. The raw `key` value is only present on **create** and **rotate** responses, so callers must capture it immediately. This matches the standard secure-by-default pattern for credential APIs: the server cannot hand the raw key back later, so a missed capture forces a rotation rather than silently leaking a long-lived secret.

## Reviewer Notes

Things worth a careful look:

- `src/kitaru/client.py` — the whole `_validate_auth_pagination` helper and the `AuthAPIKeyManager` / `AuthServiceAccountManager` classes. Check the validation rules (especially the `limit` vs `page`/`size` exclusivity and the `isinstance(..., bool)` guards, which are there because `bool` is a subclass of `int` in Python).
- `src/kitaru/_cli/_auth.py` — the activation/deactivation rollback path added in the last two commits (`Fix auth key activation handling`, `Harden auth key activation rollback`). Worth confirming the error messaging + exit codes match the rest of the CLI.
- `_map_auth_api_key_with_value` vs `_map_auth_api_key` — confirm there is no code path on list/show that accidentally routes through the `WithValue` mapper.
- `docs/content/docs/guides/authentication.mdx` — particularly the claim that PATs are Pro-only and should not be the default OSS primitive. Make sure we're happy with that framing.

### Reproduction snippet

Run against a local Kitaru server (the smoke test already does this, but a manual pass is useful):

```bash
# Start a local server and log in
kitaru login

# SDK flow
uv run python - <<'PY'
from kitaru import KitaruClient

client = KitaruClient()
sa = client.auth.service_accounts.create(
    name="ci-runner",
    description="CI runner",
)
key = client.auth.api_keys.create(
    service_account=sa.name,
    name="default",
)
print("raw value only visible here:", key.key)

# Later: list/show never includes the raw value
for k in client.auth.api_keys.list(service_account=sa.name):
    assert not hasattr(k, "key")

# Rotate returns a new raw value
rotated = client.auth.api_keys.rotate(service_account=sa.name, name="default")
print("rotated value:", rotated.key)

# Clean up
client.auth.api_keys.delete(service_account=sa.name, name="default")
client.auth.service_accounts.delete(sa.name)
PY

# CLI flow (equivalent)
kitaru auth service-accounts create ci-runner --description "CI runner"
kitaru auth api-keys create ci-runner default -o json   # capture .item.key
kitaru auth api-keys list ci-runner -o json
kitaru auth api-keys rotate ci-runner default -o json
kitaru auth api-keys delete ci-runner default
kitaru auth service-accounts delete ci-runner
```

On a Kubernetes remote stack the flow is identical — the auth surface talks to the Kitaru server, not the execution stack, so nothing stack-specific changes.

Fixes #195


docs/content/docs/guides/authentication.mdx
```diff
@@ -0,0 +1,153 @@
...
+
+Use `kitaru auth token` when a raw HTTP client needs an `Authorization: Bearer`
+header:
+
+```bash
+KITARU_SERVER_ACCESS_TOKEN="$(kitaru auth token)"
+
+curl -H "Authorization: Bearer ${KITARU_SERVER_ACCESS_TOKEN}" \
+  https://kitaru.example.com/api/v1/...
+```
```
AlexejPenner, line 106: Ahh, we should also introduce this into the frontend code snippets to make them copy-paste-able for real

AlexejPenner (dismissed):
🦭

PR #247, opened 2026-04-29
Add dependency security automation

strickvl:
## What changed

This PR adds dependency security automation **and** runs the first remediation pass.

### Automation

- **Dependabot version-update coverage** for:
  - Python deps at `/` via the `uv` ecosystem (updates both `pyproject.toml` and `uv.lock` coherently — important because CI installs from the lock via `uv sync --frozen`)
  - Docs deps at `/docs` via the `npm` ecosystem (covers pnpm lockfiles)
  - Site deps at `/site` via the `npm` ecosystem (covers pnpm lockfiles)
- **CI `audit` job** that runs `uv run pip-audit` on every PR, with a documented ignore list at `.github/pip-audit-ignored.txt` for accepted-risk CVEs.
- **`just audit` recipe** so contributors can run the same check locally.
- Reused the existing Dependabot schedule, reviewer, `dependencies` label, cooldown policy, and `develop` target branch.

### Remediation pass

`pip-audit` initially reported **24 vulnerabilities across 12 packages**. After this PR:

- **18 patched** via `uv lock --upgrade`: aiohttp, anthropic, cryptography, pygments, pytest, python-multipart, requests, and 11 more.
- **6 remaining**, all upstream-blocked and documented in `.github/pip-audit-ignored.txt` with rationale + tracking link:
  - `pyjwt` + `mcp`: blocked by zenml's `pyjwt==2.7.*` pin → escalated as **zenml-io/zenml#4782**.
  - `orjson` + 2x `starlette`: transitive zenml server-stack pins; will refresh when the keystone pin above widens.
  - `pip`: tooling, not a runtime dep.

### Knock-on changes from the lock refresh

- **Capped `pydantic-ai-slim<1.86.0`** — 1.86 added `capabilities`/`spec` kwargs to `Agent.run()` that the `KitaruAgent` adapter doesn't pass through, breaking 4 phase17 tests. Tracked as **#248** for proper adapter support.
- **Migrated `# type: ignore[X]` → `# ty: ignore[Y]`** across 7 files. The `ty 0.0.21 → 0.0.32` upgrade surfaced 23 diagnostics on lines that previously suppressed them via mypy-style comments; ty has its own syntax with codes like `invalid-argument-type`, `unknown-argument`, `unresolved-attribute`, `invalid-assignment`, and `invalid-raise`.

Closes #79.
Closes #80.

## Why

Two pieces of supply-chain hygiene plus the immediate remediation:

1. Dependabot can now raise update PRs for Python and pnpm-managed Node deps. The Python lane uses the `uv` ecosystem so PRs update `uv.lock` (the source of truth for `uv sync --frozen`), not just `pyproject.toml`.
2. CI now scans the Python environment for known vulnerabilities, gated by an explicit accepted-risk ignore file rather than silent suppression.
3. The first scan revealed 24 CVEs; this PR patches everything we control and documents what we don't.

- **Dependabot `uv` ecosystem** (not `pip`): caught by codex review — `pip` would update `pyproject.toml` constraints but leave `uv.lock` stale, so CI would silently install the old version.
- **Documented ignore list, not blanket suppression**: each ignored CVE includes a rationale and a tracking link. Removing an entry is a one-line diff once upstream resolves the cause.
- **Did not force-override transitive zenml pins** (orjson, starlette, pyjwt): the runtime breakage risk outweighs the bounded security gain, especially since these are server-side deps in a zenml-managed stack typically deployed in trusted clusters.
- **`xargs` over inline `$var` expansion** in the CI step: avoids `SC2086` shellcheck warning while keeping the ignore-list parser dead simple.

## Reviewer Notes

To reproduce locally:

```bash
# Verify the audit gate passes with the documented ignore list
just audit

# Or equivalently:
awk '/^CVE-|^GHSA-/ {printf "--ignore-vuln %s ", $1}' \
  .github/pip-audit-ignored.txt | xargs uv run pip-audit
# Expected: "No known vulnerabilities found, 6 ignored"

# Verify CI checks
just check
just test
```

Expected results:
- `just audit` → "No known vulnerabilities found, 6 ignored"
- `just check` → All checks pass
- `just test` → 1584 passed, 0 failed

Worth taking a careful look at:
- `.github/pip-audit-ignored.txt` — make sure each rationale rings true; this is where future-us will revisit.
- `pyproject.toml:62-67` — the `pydantic-ai-slim<1.86.0` cap should come back off once #248 lands.
- The `# ty: ignore` migrations across the test files — quick sanity check that the chosen ty error codes match what the upgraded `ty` actually reports.


AlexejPenner (approved):
🦭

PR #251, opened 2026-04-29
Handle runtime log version skew errors

### LINKED ISSUE #136: kitaru executions logs surfaces raw server-side ImportError on version skew
### What happened

While running through the quickstart against a remote Kitaru server (after `kitaru login <server>`), invoking `kitaru executions logs <exec_id>` returned a raw server-side `ImportError` to the client:

```
Error: Failed to retrieve runtime logs for source 'step': No module named 'zenml.log_stores'
```

The error message surfaces an internal Python module name to the user with no actionable hint about what's wrong or how to fix it. After running `kitaru logout` and returning to the local server, `executions logs` works fine (returns either `No log entries found` or, in `local database` mode, a clear `Runtime log retrieval requires a server-backed connection` message).

### What I expected

If the connected server is on a Kitaru/ZenML version that doesn't expose the log retrieval surface, the CLI should produce an actionable error along the lines of:

> The connected Kitaru server (vX.Y.Z) does not support the runtime logs API. Upgrade the server to vA.B.C+ or run `kitaru logout` to use a local connection.

Even a generic version-skew detection wrapper would be a big DX win — surfacing `No module named 'zenml.log_stores'` is a debugging dead end for users who don't know that ZenML is the underlying engine.

### Reproduction

1. `kitaru login <some-remote-server-on-older-zenml>`
2. Run any flow that produces an execution.
3. `kitaru executions logs <exec_id>`
4. Observe the raw `ImportError` text in the CLI output.

I encountered this organically (the user briefly logged into a remote server mid-session). I don't have a stable remote target on the older zenml version anymore to re-reproduce on demand, but the error path in `kitaru.client.logs` lines 385-404 (`client.py`) suggests this happens whenever the server raises any non-empty/non-OTEL exception — the message is passed through verbatim:

```python
raise KitaruLogRetrievalError(
    f\"Failed to retrieve runtime logs for source '{source}': {exc}\"
) from exc
```

So any server-side import error, attribute error, or unhandled exception in the logs endpoint will reach the user as raw exception text.

### Suggested direction

- Wrap server-side errors with a friendlier message and a hint to check server/client version compatibility.
- Optionally: add a version check on connect (`KitaruClient` could fetch server `/api/v1/info` or equivalent and warn on mismatch).
- Optionally: detect the specific `ImportError: No module named 'zenml.log_stores'` and surface a targeted message about server upgrade requirements.

### Environment

- **Kitaru (client)**: 0.4.0
- **Python**: 3.12.13
- **OS**: macOS (arm64)
- **Connection**: was logged into remote server when error surfaced; cannot reproduce locally now (returned to local server / local database mode)

strickvl:
## What changed

- Detect the known older-server runtime log endpoint failure that reports `zenml.log_stores` as a missing module.
- Translate that failure into a Kitaru-branded compatibility error instead of surfacing the raw backend import error.
- Add regression coverage for both direct `ModuleNotFoundError` and string-wrapped remote error shapes.

Fixes #136.

## Why

When `kitaru executions logs <exec_id>` talks to an older/incompatible remote server, the backend can fail with an internal import error. Previously Kitaru wrapped that exception verbatim, so users saw a confusing implementation detail instead of an actionable next step.

The new message tells users that the server log endpoint is incompatible with their client and suggests aligning/upgrading client/server versions, without mentioning backend internals.

## Reviewer Notes

Please pay special attention to:

- `src/kitaru/client.py`: the new branch in `_ExecutionsAPI._fetch_log_payload()` should stay after the empty-log and OTEL-specific handling, but before the generic fallback.
- `src/kitaru/_client/_logs.py`: `_is_log_endpoint_version_skew_error()` intentionally detects only the known leaked module path to avoid broad string matching.
- `tests/test_client.py`: the regression test checks both that the new message is actionable and that it does **not** leak `ZenML`, `zenml.log_stores`, `No module named`, or `ModuleNotFoundError`.

To reproduce the original failure path locally, run the regression test added in this PR:

```bash
uv run pytest tests/test_client.py -k logs_map_version_skew_errors_to_actionable_kitaru_error -q
```

That test simulates the remote log endpoint raising the same older-server import failure reported in #136 and verifies the CLI/SDK-facing error text stays sanitized.

## Validation

```bash
uv run pytest tests/test_client.py -k "logs_map_version_skew or logs_map_otel_retrieval_errors or logs_return_empty_list_when_backend_reports_no_entries"
uv run pytest tests/test_cli.py -k "executions_logs_empty_state or executions_logs_surfaces_backend_errors"
just check
just test
```

Results:

- Focused client log tests: 4 passed
- Focused CLI log tests: 2 passed
- `just check`: passed
- `just test`: 1586 passed, 8 skipped

### DIFF
--- a/src/kitaru/_client/_logs.py
+++ b/src/kitaru/_client/_logs.py
@@ -163,11 +163,17 @@ def _is_otel_log_retrieval_error(message: str) -> bool:
     return "otel" in lowered and "not implemented" in lowered
 
 
+def _is_log_endpoint_version_skew_error(message: str) -> bool:
+    """Return whether an error message points to incompatible log endpoints."""
+    return "zenml.log_stores" in message.lower()
+
+
 __all__ = [
     "_coerce_log_level",
     "_coerce_log_lineno",
     "_coerce_log_text",
     "_is_empty_log_result_error",
+    "_is_log_endpoint_version_skew_error",
     "_is_otel_log_retrieval_error",
     "_log_sort_key",
     "_map_runtime_log_entry",
--- a/src/kitaru/client.py
+++ b/src/kitaru/client.py
@@ -52,6 +52,7 @@ from kitaru._client._logs import (
     _coerce_log_lineno,
     _coerce_log_text,
     _is_empty_log_result_error,
+    _is_log_endpoint_version_skew_error,
     _is_otel_log_retrieval_error,
     _log_sort_key,
     _map_runtime_log_entry,
@@ -432,6 +433,16 @@ class _ExecutionsAPI:
                 if endpoint_hint:
                     message += f" View them in your OTEL backend at: {endpoint_hint}."
                 raise KitaruLogRetrievalError(message) from exc
+
+            if _is_log_endpoint_version_skew_error(error_message):
+                raise KitaruLogRetrievalError(
+                    "Unable to retrieve runtime logs because the server log "
+                    "endpoint is incompatible with this Kitaru client. This "
+                    "usually means the client and server are running different "
+                    "Kitaru versions. Upgrade the server runtime or align the "
+                    "client and server versions, then retry `kitaru executions "
+                    "logs`."
+                ) from exc
 
             raise KitaruLogRetrievalError(
                 f"Failed to retrieve runtime logs for source '{source}': {exc}"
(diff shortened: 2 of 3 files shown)


AlexejPenner (approved):
🦭

PR #289, opened 2026-05-04
Remove native memory feature

### LINKED ISSUE #312: Remove memory feature

strickvl:
## What changed

PR1 of the native-memory removal stack. This hard-removes the Kitaru-native memory surface from the core package:

- removes `kitaru.memory`
- removes `KitaruClient.memories`
- removes the `kitaru memory ...` CLI command group
- removes `kitaru_memory_*` MCP tools
- deletes the `src/kitaru/memory/` implementation package
- deletes the shared memory CLI/MCP interface helper
- removes memory-specific analytics events, error class, inspection serializers, CLI compatibility exports, and flow/checkpoint wiring
- removes the new upstream `memory_call` checkpoint mapping because memory operations no longer exist
- restores non-memory MCP `kitaru_secrets_create` tests that were accidentally swept out with the memory MCP test block

The Pydantic AI adapter's local/in-process `persist_message_history` behavior is intentionally preserved. The only Pydantic AI change is README wording so durable history points to user-owned storage rather than `kitaru.memory`.

## Why

We are removing Kitaru-native memory as a hard cut rather than deprecating it. There are not meaningful users relying on this API yet, and removing it now keeps the core product surface smaller before more docs/examples/site material accretes around it.

- Chose hard removal instead of compatibility shims: importing `kitaru.memory` from the built wheel now raises `ModuleNotFoundError`.
- Kept examples/docs/site/blog mostly untouched for follow-up stacked PRs.
- Added explicit module-level PR2 TODO skips for memory-dependent example test modules so PR1 can remove the core API without pretending the examples are already migrated:
  - `tests/test_news_scout_example.py`
  - `tests/test_phase20_memory_example.py`
  - `tests/test_phase3_compliance_review_stage3.py`
- Kept `kitaru_secrets_create` MCP coverage because secret creation is a non-memory MCP surface and should still prove it returns metadata only, forwards `private=True`, propagates SDK errors, and does not expose a delete tool.

## Deferred follow-ups

This PR intentionally leaves user-facing cleanup for stacked follow-ups:

- PR2: examples cleanup/rewrite
- PR3: hand-written docs cleanup
- PR4: website/blog/marketing/comparison cleanup

## Verification

Original full PR1 verification:

```bash
just fix
just check
just test
just build
```

Original full test result:

```text
1460 passed, 11 skipped, 82 warnings in 27.04s
```

Additional review-fix verification for restored MCP secret tests:

```bash
uv run pytest tests/mcp/test_server.py -k 'secrets_create or secret_delete'
uv run pytest tests/mcp/test_server.py
just check
```

Results:

```text
4 passed
124 passed
All checks passed
```

Wheel smoke also passed:

```bash
uv run --isolated --with dist/kitaru-0.7.0-py3-none-any.whl -- python - <<'PY'
import importlib
import kitaru
importlib.import_module('kitaru.memory')
PY
```

Expected behavior confirmed: `kitaru.memory` is absent from the built wheel, and `kitaru --help` from the wheel no longer lists a memory command.

## Reviewer Notes

Please focus review on the core public-surface removal rather than examples/docs/site copy, which are intentionally deferred.

Useful checks:

```bash
# Confirm removed public API strings are gone from source/tests.
git grep -n -E 'kitaru\.memory|from kitaru import memory|\.memories\b|memory_app|kitaru_memory_|memory_call|MEMORY_CALL' -- src tests ':!src/kitaru/adapters/pydantic_ai/README.md'

# Confirm the wheel does not expose memory.
rm -rf dist build
just build
wheel=$(realpath "$(ls -1 dist/kitaru-*.whl | head -1)")
cd /tmp
uv run --isolated --with "$wheel" -- python - <<'PY'
import importlib
import kitaru
try:
    importlib.import_module('kitaru.memory')
except ModuleNotFoundError:
    print('kitaru.memory absent as expected')
else:
    raise SystemExit('kitaru.memory unexpectedly importable')
PY
```

Also check that `persist_message_history=True` in the Pydantic AI adapter remains intact; this PR removes Kitaru-native durable memory, not local adapter message history.


AlexejPenner (dismissed):
🦭

PR #305, opened 2026-05-05
Make `.wait()` ambiguity discoverable + recommend `runner_call` for OpenAI Agents adapter

strickvl:
## Summary

While running pre-release smoke tests for v0.9.0, the OpenAI Agents adapter integration example crashed on `.wait()` with a `KitaruRuntimeError` whose message said:

> Execution output metadata is missing and fallback extraction is ambiguous because multiple terminal steps were found.

…and gave the user no way to know that the per-call artifacts were still in the Kitaru UI, or how to retrieve them, or what alternative to try. The headline new feature of v0.9.0 was failing on its own documented "Minimal flow" example.

This PR keeps the loud raise (silencing it would hide the real per-checkpoint artifacts from users — a quiet observability regression) and makes the raise *actionable*. It also flips the documented adapter default from `calls` → `runner_call`, since that's the strategy that produces a clean single-sink terminal artifact and lets `.wait()` return a value cleanly. Several follow-on consistency fixes also land here: a stale README, a CHANGELOG typo, a doc workaround that contradicted an adapter guardrail, the gated research-bot smoke test (which had the same ambiguity bug pre-existing), and the research-bot CLI surfacing the new error humanely instead of as a stack trace.

### Core (`src/kitaru/`)
- **New `KitaruAmbiguousFlowResultError`** (subclass of `KitaruRuntimeError`) raised by `_extract_values_from_terminal_steps` in `src/kitaru/flow.py`. Callers can now target the ambiguity case specifically — the old `except KitaruRuntimeError` would also catch `KitaruExecutionError` (real run failures) and silently swallow them. Re-exported from `kitaru.errors` and the package root.
- **New error message** names the colliding terminal checkpoints (or the multiple outputs of a single terminal, by name), gives the execution ID, points at the Kitaru UI + `KitaruClient().executions.get('<id>')`, and suggests `checkpoint_strategy="runner_call"` as the alternative.

### OpenAI Agents adapter docs + examples
- **`docs/content/docs/guides/openai-agents-adapter.mdx`**: leads with `runner_call` as the documented default; first install snippet uses `runner_call` explicitly. Drops a workaround sentence that contradicted the adapter's `_require_calls_scope()` guardrail (wrapping a `calls` runner in your own `@checkpoint` raises — the docs now say so).
- **`examples/integrations/openai_agents_agent/openai_agents_adapter.py`**: defaults to `runner_call`. `calls` becomes an opt-in comparison (`OPENAI_AGENTS_COMPARE_CALLS=1`, renamed from `OPENAI_AGENTS_COMPARE_RUNNER_CALL`), with an explicit `try/except KitaruAmbiguousFlowResultError` that *demonstrates* what `calls` strategy does to `.wait()` instead of hiding it.
- **`examples/integrations/openai_agents_agent/README.md`** updated to match the new default and the renamed env var (catch from Codex review).
- **`examples/end_to_end/openai_research_bot/research_bot.py`**: the CLI's `_run_once` now catches `KitaruAmbiguousFlowResultError` and prints a friendly summary (execution URL, `KitaruClient` pointer, the full error text) instead of letting the user see a stack trace when they pick `--strategy calls`. The research bot's `publish_report` checkpoint already threads the side-effect artifacts to be a single sink under `runner_call`, but the adapter's per-call peer checkpoints under `calls` cannot be threaded — so this CLI ambiguity was a pre-existing bug that the new error message exposed.

### Smoke test
- **`scripts/smoke-test.sh`**: gated `KITARU_SMOKE_RESEARCH_BOT=1` entry switched from `--strategy calls` to `--strategy runner_call`. Validates the documented happy path on release rehearsals.

### CHANGELOG
- Fixed pre-existing `KitaruAgent` → `KitaruRunner` typo in the v0.9.0 OpenAI Agents adapter bullet.
- Added "Changed" section noting the new dedicated error class and the recommendation flip.

## Reviewer Notes

The most important pieces to look at:

- **`src/kitaru/errors.py`** — the new exception class. Subclassing `KitaruRuntimeError` keeps existing `except KitaruRuntimeError` callers compatible while letting newer code be more specific.
- **`src/kitaru/flow.py`** — the new `_ambiguous_terminal_message(execution_id, *, reason)` helper and the two raise sites. The single-terminal-multiple-outputs branch now lists the output names too (was previously a count only).
- **`examples/integrations/openai_agents_agent/openai_agents_adapter.py`** — the `try/except KitaruAmbiguousFlowResultError` block is intentionally visible in the example body. The comment explains *why* (per-call peers, no single sink), which is the educational point. Examples stay as primary teaching surface, not behind helpers.
- **`examples/end_to_end/openai_research_bot/research_bot.py`** — same pattern at the CLI's `_run_once`, with the friendly summary text giving users the URL + `KitaruClient` next-step.
- **`docs/content/docs/guides/openai-agents-adapter.mdx`** — note the "Important guardrail" interaction: the section now warns explicitly that wrapping the `calls` runner in your own `@checkpoint` is **not** a workaround.

### How to reproduce

**Locally (the original failure on `develop` before this PR):** with `OPENAI_API_KEY` set:

```bash
uv sync --extra local --extra llm --extra mcp --extra openai-agents
uv run kitaru init
uv run examples/integrations/openai_agents_agent/openai_agents_adapter.py
```

You'll see the run complete server-side, then `.wait()` raise the unhelpful "fallback extraction is ambiguous" message.

**With this PR:** the same command runs `runner_call` strategy by default and prints the model output cleanly. To exercise the (now opt-in) `calls` path and see the new actionable error message:

```bash
OPENAI_AGENTS_COMPARE_CALLS=1 uv run examples/integrations/openai_agents_agent/openai_agents_adapter.py
```

You'll see the `runner_call` output, then a `=== calls strategy output ===` section showing the new message that lists the terminal checkpoints, gives the execution ID, and suggests `KitaruClient` + `runner_call`.

**Research bot CLI (graceful ambiguity handling):**

```bash
uv run python examples/end_to_end/openai_research_bot/research_bot.py \
  "AI agent durability in one paragraph" --max-searches 2 --strategy calls --fail-on-search-error
```

Exits 0 with a friendly summary pointing at the execution URL, instead of a stack trace.

**Gated smoke test (pre-release rehearsal of the recommended path):**

```bash
KITARU_SMOKE_RESEARCH_BOT=1 ./scripts/smoke-test.sh -s
```

Now exercises `--strategy runner_call` and exits with `Failed: 0`.

**On a remote stack (k8s):** the same examples work against a remote stack — `runner_call` is what users will want by default there too, since they typically `.wait()` and consume the return value.

### Test plan

- [x] `just lint` — clean
- [x] `just typecheck` — clean
- [x] `just format-check` — clean
- [x] `just test` — 1731 passed, 8 skipped (added 3 new tests: error message contents, output name listing, error class hierarchy)
- [x] Manual reproduction of integration example shows the new error message text
- [x] Research bot CLI surfaces friendly summary instead of stack trace
- [x] Confirm gated smoke test passes end-to-end with `KITARU_SMOKE_RESEARCH_BOT=1` (recommended pre-merge sanity check)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,9 +8,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ## [Unreleased]
 
 ### Added
-- OpenAI Agents SDK adapter (`kitaru.adapters.openai_agents`) — wrap an `Agent`/`Runner` with `KitaruAgent` to make OpenAI Agents SDK runs durable, replayable, and observable under a Kitaru flow. Supports two tracking strategies via `checkpoint_strategy="calls"` (per-tool/per-model checkpoints) or `checkpoint_strategy="runner_call"` (one checkpoint per `Runner.run`), plus a guide page at `/guides/openai-agents-adapter` that walks through the trade-offs. (#295)
+- OpenAI Agents SDK adapter (`kitaru.adapters.openai_agents`) — wrap an `Agent`/`Runner` with `KitaruRunner` to make OpenAI Agents SDK runs durable, replayable, and observable under a Kitaru flow. Supports two tracking strategies via `checkpoint_strategy="runner_call"` (one checkpoint per `Runner.run`, recommended when you want a clean `.wait()` return value) or `checkpoint_strategy="calls"` (per-tool/per-model checkpoints for finer replay units, with per-checkpoint artifacts visible in the Kitaru UI / `KitaruClient`). The guide at `/guides/openai-agents-adapter` walks through the trade-offs. (#295)
 - OpenAI Agents integration example (`examples/integrations/openai_agents_agent/`) and an end-to-end `openai_research_bot` example (planner + parallel search checkpoints + writer report generation, with remote secret guidance and Kitaru UI artifacts). Both are exercised by the smoke test. (#295)
 - Markdown exports for every docs page at `kitaru.ai/docs/<slug>.md`, plus a substantially expanded `/llms.txt` index — making the docs friendlier for LLMs and agents that consume them programmatically. (#303)
+
+### Changed
+- `flow.run(...).wait()` now raises a new dedicated `KitaruAmbiguousFlowResultError` (subclass of `KitaruRuntimeError`) when the flow has multiple terminal checkpoints with no single sink (common with the OpenAI Agents adapter's `checkpoint_strategy="calls"`). The error names the terminal checkpoints, points at the execution's artifacts in the Kitaru UI, and suggests `KitaruClient` retrieval and the `runner_call` strategy as alternatives. Catching this specific subclass lets callers handle the ambiguity case without accidentally swallowing real execution failures.
 
 ## [0.8.0] - 2026-05-04
 
--- a/docs/content/docs/guides/openai-agents-adapter.mdx
+++ b/docs/content/docs/guides/openai-agents-adapter.mdx
@@ -11,8 +11,12 @@ from agents import Agent
 from kitaru.adapters.openai_agents import KitaruRunner
 
 agent = Agent(name="researcher", model=your_model)
-runner = KitaruRunner(agent)
+runner = KitaruRunner(agent, checkpoint_strategy="runner_call")
 ```
+
+The runtime default is `checkpoint_strategy="calls"` (per-call checkpoints —
+see below); pass `"runner_call"` whenever you want a single terminal
+checkpoint so `flow.run(...).wait()` returns the run result directly.
 
 You run the agent through `runner.run(...)` or `runner.run_sync(...)` with an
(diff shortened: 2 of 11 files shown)


bastizol-wq:
Re: “Make `.wait()` ambiguity discoverable + recommend `runner_call` for OpenAI Agents adapter”. When agents call tools, the brittle part is usually the tool boundary (auth/hosts/inputs).

We’ve had good luck turning this into a small policy step: validate tool description + transport, then only allow the call if checks pass.

ATCP is an API-first gate that evaluates a tool/OpenAPI and returns a structured allow/warn/block result.

If you want, I can share the exact first checks we run (host policies + auth expectations + dangerous ops).

PR #316, opened 2026-05-08
Add Claude Agent SDK adapter

### LINKED ISSUE #276: Add integration into Claude Agent SDK

strickvl:
## What changed

Adds a new `kitaru.adapters.claude_agent_sdk` adapter that wraps one Claude Agent SDK `query(...)` invocation as one Kitaru checkpoint.

The adapter captures the durable boundary payload around that invocation:

- Claude session ID
- final text/result
- SDK-emitted messages
- optional local transcript payload/path when available
- usage/model usage/cost fields when exposed by the SDK
- redacted options manifest
- event log and run summary artifact names

It explicitly rejects granular strategies such as `calls`, `runner_call`, `model_call`, and `tool_call` for Claude, because Claude hooks/messages are observable state but not safe Kitaru replay boundaries for Claude-internal model/tool/Bash/MCP side effects.

Also adds:

- Claude adapter analytics events
- focused fake-SDK tests
- an educational runnable example under `examples/integrations/claude_agent_sdk_agent/`
- docs guide at `/guides/claude-agent-sdk-adapter`
- examples index/docs navigation updates
- smoke-test wiring with safe default skip for real Claude runs unless credentials/provider mode are configured
- changelog entry

## Why it was needed

Claude Agent SDK users should be able to put existing Claude SDK work under Kitaru-managed execution while keeping the durability boundary precise.

The useful promise here is invocation-level durability, auditability, session capture, cost/usage visibility, and replay-skip for completed Claude invocations. Kitaru does not snapshot arbitrary workspaces or make Bash, MCP, custom tool, file edit, or external API side effects deterministic inside a Claude invocation.

- `checkpoint_strategy="invocation"` is the only supported strategy.
- `KitaruClaudeRunner` requires a stable name and should normally be called from the Kitaru flow body so the adapter can create its own invocation checkpoint.
- Calls from inside an existing Kitaru checkpoint are rejected by default. Users can opt in with `allow_direct_execution_inside_checkpoint=True`, but replaying the outer checkpoint can call Claude again.
- `options_factory` is preferred for request-scoped Claude options like `cwd`, `resume`, and `max_turns`.
- Transcript-file capture is best-effort; SDK-emitted messages are the primary captured message artifact.
- Artifact capture and event/log persistence are best-effort by default after Claude succeeds, so observability storage hiccups do not turn a completed Claude invocation into a retry. Strict behavior is available with `ClaudeCapturePolicy(fail_on_artifact_capture_error=True, fail_on_event_persistence_error=True)`.
- Options manifests are redacted by default, including common key/value sequence shapes such as header pairs.
- The example and docs avoid streaming APIs because Kitaru/ZenML streaming support is not available yet.
- The first integration example is deliberately educational and tool-free so users can understand the one-invocation/one-checkpoint boundary before adding Bash/MCP/custom tools.

### Follow-up docs polish

After opening the PR, the Claude adapter guide was expanded to tell the fuller story for existing Claude Agent SDK users: how the Kitaru boundary fits around `query(...)`, how sessions differ from Kitaru replay, what artifacts are captured, how `options_factory` should be used, how Claude tools/MCP/Bash/hooks remain inside the invocation boundary, and where to read the official Claude SDK docs.

The guide was later simplified to avoid adapter-to-adapter comparison language and version-promise framing, so the page describes the Claude adapter directly on its own terms.

### Review follow-up

Addressed the transcript-capture review feedback:

- Transcript lookup now uses the effective Claude SDK working directory: request `cwd`, then `options.cwd` / mapping `cwd`, then the current process directory.
- Claude `session_id` values are validated before being used as transcript path components, and resolved transcript paths must stay under Claude's expected transcript directory.
- The docs now explain the Kubernetes / distributed-orchestrator boundary: completed Kitaru checkpoint outputs and captured artifacts are durable through Kitaru storage, while local Claude JSONL transcript files and workspaces are pod-local unless backed by shared storage or Claude session-store support.

Addressed final replay-safety review feedback:

- Nested checkpoint execution is now rejected by default before the Claude SDK is called.
- Direct nested execution requires explicit opt-in and records a warning/metadata flag.
- Artifact capture failures are best-effort by default and reported in `result.warnings` / `result.metadata["capture_failures"]`.
- Event/log persistence failures are best-effort by default and reported in `result.warnings` / `result.metadata["event_persistence_failures"]`.
- Strict capture/persistence failure behavior is available through `ClaudeCapturePolicy`.
- Options-manifest redaction now handles header/env-style key/value pairs such as `[("Authorization", "Bearer ...")]`.
- The Claude adapter example README now emphasizes the recommended flow-body pattern and warning output.

## Reviewer Notes

Please focus review on the exact durability claim and public API shape:

- `src/kitaru/adapters/claude_agent_sdk/_agent.py`
- `src/kitaru/adapters/claude_agent_sdk/_runner.py`
- `src/kitaru/adapters/claude_agent_sdk/_serialization.py`
- `src/kitaru/adapters/claude_agent_sdk/_tracking.py`
- `docs/content/docs/guides/claude-agent-sdk-adapter.mdx`
- `examples/integrations/claude_agent_sdk_agent/claude_agent_sdk_adapter.py`

Useful checks:

```bash
uv sync --extra claude-agent-sdk
uv run pytest -n 0 \
  tests/test_claude_agent_sdk_adapter_foundation.py \
  tests/test_claude_agent_sdk_invocation_strategy.py
uv run python examples/integrations/claude_agent_sdk_agent/claude_agent_sdk_adapter.py --help
```

Latest local verification:

- `just fix` ✅
- `just check` ✅
- Claude adapter focused tests ✅ `48 passed`
- `just test` ✅ `1653 passed, 8 skipped`

### Final options-handling follow-up

Addressed the last final-review items:

- Synthetic checkpoint cache keys now use deterministic structural option identities instead of live-object reprs / memory-address-like strings.
- Options-manifest construction failures are treated as artifact-capture failures by default and no longer mask completed Claude calls or original Claude SDK errors.
- Options-manifest redaction now handles env/header entries shaped like `{"name": "ANTHROPIC_API_KEY", "value": "..."}`.
- Transcript lookup now prefers the actual SDK options `cwd` over `ClaudeRunRequest.cwd`, so `options_factory` cwd overrides resolve the local Claude JSONL transcript from the directory Claude really used.

Latest local verification after this follow-up:

- `just fix` ✅
- `just check` ✅
- Claude adapter focused tests ✅ `54 passed`
- `just test` ✅ `1659 passed, 8 skipped`


strickvl:
FYI @htahir1

strickvl:
Addressed the transcript-capture review feedback in af2fd58:\n\n- Transcript discovery now uses the effective Claude SDK cwd: request cwd, then options cwd, then process cwd.\n- Session IDs are validated before being used in transcript paths, with a resolved-path containment check.\n- The Claude guide now calls out the Kubernetes/distributed-runner boundary: captured Kitaru artifacts are durable, but local Claude JSONL files/workspaces are only pod-local unless shared storage or session-store support is configured.\n\nVerification after the fix: focused Claude adapter tests pass (38 passed), full test suite passes (1836 passed, 8 skipped), and just check still only fails at the pre-existing lychee.toml parse issue.

htahir1:
@strickvl maybe here we can use hooks in Anthropic Agents SDK to do more finer grained checkpointing?

strickvl:
> @strickvl maybe here we can use hooks in Anthropic Agents SDK to do more finer grained checkpointing?

@htahir1 do you want this for replay + durability, or just for the UI having things to display etc?

strickvl:
FWIW @htahir1 I deliberately didn't get too much into messing with hooks in this first version for the reasons described below (derived from the plans I put together):

Good question. Hooks are definitely the right seam to look at for observability, policy, approvals, and maybe future audit trails. We did discuss that path while shaping this adapter.

The reason I’m hesitant to use hooks as Kitaru checkpoints is that it would create a replay promise we cannot honestly keep yet.

The current adapter boundary is deliberately outside the Claude loop: from flow scope, `KitaruClaudeRunner` opens one synthetic `<runner>_claude_invocation` checkpoint, calls `claude_agent_sdk.query(...)`, drains messages until the final `ResultMessage`, and stores a serializable `ClaudeRunResult` plus artifacts like usage, final output, transcript data when available, redacted options, event log, and summary. On replay, Kitaru can safely say: “this completed Claude invocation already happened, so reuse its completed output.”

Hooks sit inside that Claude invocation. They can tell us things like “Claude is about to call Bash” or “Claude just used this MCP tool”, and in some cases they can block, modify, or add context. But that is not the same as having a replay-safe checkpoint for that internal step.

The bad user experience would look like this:

1. Claude runs a tool or Bash command inside the invocation.
2. A hook fires, so Kitaru records “tool call happened.”
3. The tool writes a file, mutates a workspace, calls an external API, posts to Slack, updates a database, or triggers some MCP-side effect.
4. The run crashes before the outer Claude invocation finishes.
5. On replay, Kitaru has no complete, deterministic checkpoint output for that hook boundary.

At that point we have two bad choices:
- run Claude/tooling again and risk repeating the side effect;
- or pretend we can skip forward from the hook observation, even though we may not have the filesystem state, external state, Claude internal loop state, permission decisions, or tool output needed to reconstruct the world.

That would be a pretty confusing experience for users. They would see “fine-grained checkpointing” and reasonably assume “Kitaru will not redo this expensive/dangerous thing,” but under the hood we would only have observed that the thing happened — not captured enough state to safely replay around it.

Claude sessions, transcripts, hooks, and Claude’s file checkpointing are all useful, but they are not the same as a completed Kitaru checkpoint. In particular, Claude sessions preserve conversation history, not the filesystem; and Claude file checkpointing is limited to certain file-editing tools, not arbitrary Bash/MCP/custom tool/database/network side effects.

That is why this PR intentionally keeps the public contract at `checkpoint_strategy="invocation"`. The code rejects `calls`, `model_call`, `tool_call`, `granular`, etc.; the event model is invocation-only; and `runtime="isolated"` is rejected because Claude options can include live hooks/callbacks/MCP objects that Kitaru cannot reconstruct across an isolated checkpoint boundary.

I do think hooks are a good future extension point, just not as checkpoints themselves. More honest uses would be:

- tracing / child-event visualization;
- audit logs of tool calls and permission decisions;
- side-effect ledgers;
- policy enforcement before risky tools run;
- Kitaru-owned wrappers around specific tools/MCP/Bash surfaces where we can add idempotency and artifact capture;
- pairing Claude session/transcript data with Kitaru-owned workspace snapshot or volume semantics.

So my take is: keep this PR at invocation-level durability. That is the safe replay claim we can make today. If we want finer-grained Claude durability later, I’d rather build explicit Kitaru-owned side-effect boundaries than label hook observations as checkpoints.

PR #318, opened 2026-05-09
Add checkpoint and OpenAI agent streaming

strickvl:
## What changed

This draft PR adds checkpoint-level streaming support to Kitaru, plus OpenAI Agents SDK streaming support on top of it.

The main pieces are:

- A public `kitaru.events` surface for publishing custom checkpoint events.
- Checkpoint progress/custom event publishing and stream watching support.
- Client-side event models and stream consumption helpers.
- CLI/example watcher scripts for watching execution events from a terminal.
- OpenAI Agents SDK streaming support through `KitaruRunner.run_stream(...)` / `run_stream_sync(...)`.
- OpenAI stream event normalization so text deltas can be consumed by UIs.
- Docs and examples for checkpoint streaming and OpenAI Agents streaming.
- A small real OpenAI live-events showcase flow for demoing the UI with a longer Aria/Blupus story stream.

This depends on the ZenML streaming work from the companion feature branch. This PR does not mention any specific test endpoint or staging server; it is written as a Kitaru-side POC against streaming-enabled ZenML.

## Why

A normal completed flow tells users what happened after the fact. Streaming lets Kitaru tell the user what is happening while the flow is still alive: a checkpoint started, progress moved forward, a meaningful custom event was emitted, or an agent is writing its answer token by token.

For the OpenAI Agents SDK specifically, this gives us the first end-to-end path where model output can flow through Kitaru and into a terminal watcher or UI panel as it is generated.

## Key implementation choices

- **Start with checkpoint-level streaming.** This gives users useful live feedback without trying to stream every internal operation in every integration at once.
- **Keep OpenAI streaming on `runner_call` only for now.** Per-call streaming with the `calls` strategy needs replayable per-call stream transcripts; this PR defers that deliberately instead of creating a fragile abstraction.
- **Normalize stream events before publishing.** The OpenAI Agents SDK emits SDK-shaped stream items; Kitaru publishes stable Kitaru-shaped event payloads that clients and UIs can consume.
- **Disable caching in streaming demos.** A streaming demo that immediately hits cache looks broken in a screen recording. The examples make repeat runs produce visible events.
- **Use educational examples, not clever abstractions.** The examples are written in the style we would expect users to copy: direct, readable, and explicit about what each piece demonstrates.

## Reviewer Notes

Please focus review on:

1. The public event API shape in `kitaru.events` and whether the names/payloads feel stable enough for a first POC.
2. The OpenAI streaming boundary: `runner_call` only, with `calls` streaming explicitly deferred.
3. Event serialization and client model compatibility with the streaming-enabled ZenML branch.
4. The docs tone: they should teach when to use progress events, custom events, and OpenAI text streaming, not just list APIs.
5. The examples, especially whether they are simple enough for users to copy.

A terminal demo with a real OpenAI key:

```bash
cd examples/features/checkpoint_streaming
export OPENAI_API_KEY='sk-...'

CHECKPOINT_STREAMING_DELAY_SECONDS=1.5 \
OPENAI_SHOWCASE_MIN_WORDS=320 \
uv run python ui_live_events_showcase_flow.py
```

The flow prints an execution ID. Open that execution in a streaming-aware UI, or use the watcher scripts from this PR to follow events from another terminal.

## Validation

Run during the branch work:

```bash
uv run pytest tests/test_checkpoint_streaming.py \
  tests/test_checkpoint_streaming_example.py \
  tests/test_openai_agents_streaming.py

uv run python -m py_compile \
  examples/features/checkpoint_streaming/checkpoint_streaming_flow.py \
  examples/features/checkpoint_streaming/ui_live_events_showcase_flow.py
```

The final showcase syntax check was rerun after the Aria/Blupus demo edits.


strickvl:
Closing this in favour of a separate implementation (incoming soon)

PR #322, opened 2026-05-11
Add Kitaru vs Hatchet comparison page

Tanish2697:
## Summary

Adds the 8th `/compare` landing page: **Kitaru vs Hatchet**, positioning Kitaru as a Python-first agent runtime against Hatchet's broader orchestration platform for tasks, workflows, and parallel workloads.

Page renders at `/compare/kitaru-vs-hatchet`.

## What's in the page

- 3 feature blocks with inline Astro graphics (no PNGs):
  - Runtime primitive vs orchestration platform
  - LLM calls as first-class lineage
  - Replay with checkpoint-output overrides
- `<WhenToUseEach />` (4 + 4 bullets), `<PullQuote />`
- `<ComparisonTable />` — 11 rows, balance: 3 shared / 5 Kitaru-only / 3 Hatchet-only
- "How the two surfaces map" markdown table — 8 rows mapping `hatchet.workflow`, `@durable_task`, `aio_wait_for_event`, etc. to Kitaru's `@flow`, `@checkpoint`, `kitaru.wait`, etc.
- `<CodeCompare />` with real Hatchet Python SDK code pulled from `docs.hatchet.run` (durable-event-waits, DAGs, your-first-task)
- `<ComparisonCta />` close

## Drive-by change

`site/src/components/compare/CodeCompare.astro` — removed the "Recommended" pill from the Kitaru pane. The editorial framing conflicted with the rest of the comparison-page tone (honest scoring, balanced trade-offs). Affects all `/compare/*` pages, not just Hatchet.

## OG image

Built locally as 1200×627 PNG, AVIF-compressed (5.7 KB), uploaded to R2: `https://assets.kitaru.ai/compare/a2e94421/og-card.avif`

## Reviewer Notes

A few specific things worth a second eye:

- **Hatchet code snippet** — the `hatchet.workflow(...).task(parents=[...])` + `durable_task(parents=[...])` mixed pattern in `<CodeCompare />` was assembled from verified docs but not test-run. A Hatchet user would have the strongest read on whether mixing `@task` and `@durable_task` in one DAG workflow is idiomatic, or whether this should collapse to a single durable_task with internal step memoization. Trigger API used is `hatchet.event.push(...)`.
- **Memory cleanup** — the page predates the `kitaru.memory` removal that landed on `develop` while this PR was in progress. I scrubbed all memory references and replaced with `kitaru.save()` / `kitaru.load()` where the cross-run-state story still applies. Worth a final eye on the Surface Map "Cross-run state" row and the ComparisonTable row 4 to make sure nothing slipped through.
- **`order: 80`** — picks up after Inngest at 70. Appears last on the `/compare` index until a higher-order page lands.
- **CSS duplication across the 3 new graphics** — `/simplify` flagged that the `.fig`/`.panel`/`.panel-head`/`.foot` styles repeat verbatim across all three new files. This is consistent with the existing pattern across `temporal/`, `dbos/`, `restate/`, `inngest/`, etc. Per the `kitaru-vs-page-creation` skill, extracting a shared `_graphic-primitives.css` is deferred until there's a forcing function — flagging for awareness only.
- **`/simplify` cleanup commit (`f0bcbc1`)** — dropped an orphaned `.codecompare-pill` CSS rule, removed an unused `.span--task` modifier, and ran `svgo` on `hatchet.svg` (6.4 KB → 2.3 KB, visual output unchanged).

## Test plan

- [ ] Open the Cloudflare PR preview at `/compare/kitaru-vs-hatchet` — verify the page renders end-to-end
- [ ] Check `/compare` index — verify the Hatchet card (logo, title, description) renders at the right `order: 80` position
- [ ] Click through CTA buttons
- [ ] Verify OG image still serves: `curl -sI https://assets.kitaru.ai/compare/a2e94421/og-card.avif` → HTTP 200
- [ ] Visual check that the "Recommended" pill is gone from the Kitaru pane on all comparison pages (drive-by change)
- [ ] Mobile breakpoint (< 560 px) — verify the 3 graphics collapse from side-by-side to stacked


site/public/compare/hatchet.svg
strickvl, line 1: Should this be here or in R2?

Tanish2697: I ran it by Claude @strickvl - Following the existing convention — every /compare/* page on develop (temporal, dbos, restate, claude-agent-sdk, langgraph-deep-agents, pydantic-ai) has its competitor logo at site/public/compare/<name>.svg, referenced via competitorLogo: "/compare/<name>.svg" in the MDX frontmatter. It's 2.3 KB after the svgo pass, so size isn't a concern. R2 is used for the OG card (uploaded in this PR) but not for the small inline logos.

If you want these logos to be in R2, when all compare pages are done, we can do this. Let me know

strickvl: Ok we can address this at the end.

strickvl (approved):
LGTM.

PR #334, opened 2026-05-13
Add Replay Lab prototype

strickvl:
## What changed

This PR adds a backend-first **Replay Lab** prototype for Kitaru.

The story is Sarah's support-agent regression investigation: customers say the latest agent feels worse, finance sees cost spikes, and Sarah needs more than a dashboard. She needs to replay the cases that hurt, compare a cheaper candidate, and see whether the candidate actually helps without hiding behind replay artifacts.

The prototype adds:

- an internal Replay Lab comparison engine in `src/kitaru/_replay_lab.py`
- a `kitaru_replay_lab_compare` MCP tool so Claude Code can operate the workflow through Kitaru MCP
- local `.claude/skills` for investigation, comparison, and report-reading flows
- a deterministic end-to-end demo under `examples/end_to_end/replay_lab/`
- a richer synthetic production-history seeder that creates a 12-case deterministic support history by default
- JSON, Markdown, and static HTML report generation for the demo
- tests for the comparison engine, MCP boundary, demo scenario generation, manifest compatibility, and report rendering
- a smoke-test entry for the Replay Lab example help path

## Why

The core product idea is that observability can tell Sarah what happened, but Kitaru can help her act on it.

The important measurement model is three lanes, not two:

1. **Observed production** — what originally happened.
2. **Baseline replay** — the original deployment replayed under the replay harness.
3. **Candidate replay** — the candidate replayed under the same harness.

That baseline replay lane is the control. If observed production cost $0.42, candidate replay costs $0.27, but baseline replay also costs $0.29, then most of the apparent improvement came from the replay harness rather than the candidate. The report makes that distinction visible.

## How to try it

Start with the example README:

- `examples/end_to_end/replay_lab/README.md`

The main demo path is:

```bash
uv sync --extra local
uv run kitaru login
uv run kitaru init
uv run python examples/end_to_end/replay_lab/seed_observed.py
uv run python examples/end_to_end/replay_lab/run_replay_lab.py
uv run python examples/end_to_end/replay_lab/render_report.py
```

The default seeder creates a deterministic 12-case production-like history. For the old quick path, use:

```bash
uv run python examples/end_to_end/replay_lab/seed_observed.py --small
```

Generated manifests and reports are intentionally ignored by git.

- Keep the first version backend/example/skill-first; no frontend route yet.
- Keep the example deterministic and cheap by default; no live LLM calls are required.
- Use manifest-backed cohorts as the v0 “replay pack” shape instead of adding a persistent pack registry now.
- Keep candidate behavior generic: Replay Lab passes candidate `flow_inputs` / checkpoint overrides into replay without understanding the support-agent internals.
- Keep richer synthetic history reconstructible from stable case IDs like `regulated-medical-claim--hist-02`, so replay can rebuild the same scenario later from the manifest.
- Keep richer manifest metadata as flat string labels to avoid changing the backend manifest schema.

## Reviewer Notes

Please focus review on:

1. `examples/end_to_end/replay_lab/README.md` — does the Sarah story and command path make the workflow understandable?
2. `src/kitaru/_replay_lab.py` — does the three-lane comparison model feel like the right backend seam?
3. `src/kitaru/mcp/server.py` — is the MCP tool shape suitable for Claude Code / agent-driven operation?
4. `examples/end_to_end/replay_lab/scenarios.py` and `seed_observed.py` — does the richer deterministic history feel useful without becoming too fake or too complex?
5. `.claude/skills/kitaru-replay-lab*/SKILL.md` — are these the right local skills to ship with the prototype?

Useful verification commands:

```bash
uv run pytest tests/test_replay_lab.py tests/test_replay_lab_example.py tests/mcp/test_server.py -q
uv run ruff check examples/end_to_end/replay_lab/scenarios.py examples/end_to_end/replay_lab/seed_observed.py tests/test_replay_lab_example.py
```

Optional manual demo:

```bash
uv run python examples/end_to_end/replay_lab/seed_observed.py --small
uv run python examples/end_to_end/replay_lab/run_replay_lab.py
uv run python examples/end_to_end/replay_lab/render_report.py
```

## Test coverage run locally

```bash
uv run pytest tests/test_replay_lab.py tests/test_replay_lab_example.py tests/mcp/test_server.py -q
# 139 passed

uv run pytest tests/test_replay_lab_example.py tests/test_replay_lab.py -q
# 21 passed

uv run ruff check examples/end_to_end/replay_lab/scenarios.py examples/end_to_end/replay_lab/seed_observed.py tests/test_replay_lab_example.py
# All checks passed
```


strickvl:
@AlexejPenner @htahir1 check out this little prototype and make sure to watch the walkthrough https://www.loom.com/share/9ebda8ad24ec4b389dee36d3ceaa364c video I shot. Derived heavily from this https://www.notion.so/zenml/Kitaru-Feature-Reframing-35ef8dff253880f2b9b8ef9f2fa310f2?source=copy_link#35ff8dff253880cd8f34e8a93e06f034

htahir1:
Can you add a note how this can be integrated into my agent factory (soon to be agent platform) hero example?

htahir1:
Otherwise if i get  it right this is a batch flow of running many different experiments?

strickvl:
> Otherwise if i get it right this is a batch flow of running many different experiments?

@htahir1 well the exact implementation I guess isn't so important. and in reality we'd be comparing a series of deployments (so we'd invoke against those deployments) but it's just static flows here for the sake of simplicity. But as to your bigger question:

Agent Factory shows how to build and run a production-shaped agent platform: durable agents, sandboxing, skills, credential isolation, typed services, and HITL.

Replay Lab is what you add once that platform has real executions. The factory produces durable traces (i.e. through the executions with their metadata); Replay Lab turns selected traces into a regression cohort. So if the support agent gets expensive, loops through tools, or customers say the latest skill/model change made answers worse, you can replay those cases and compare:

1. observed production
2. baseline replay
3. candidate replay

That baseline replay lane is important because it tells us whether the replay harness itself changed the measurement. So I wouldn’t describe this only as “batch experiments”; mechanically it runs a batch, but product-wise it’s a regression lab / quality gate for agents built by the factory.

The extra Agent Factory chapter could be something like “Your agents need a regression lab”: take the Stage 5/6 agent, seed or collect a production-like history, change the Profile / model / skill / service behavior, then use Replay Lab to decide whether the candidate is safe to ship. The output is a report that can be attached to the PR.

strickvl:
Shipped in #469. This was just a prototype.

PR #335, opened 2026-05-14
Expose current execution ID helper

### LINKED ISSUE #286: Expose a public helper for the current execution ID
## Summary

The `agent_factory` example needs the current flow's execution ID to name per-run Docker resources (sandbox container, proxy container, mock-services container, workspace volume). Today it imports the private `from kitaru.runtime import _get_current_execution_id`. We should expose a stable public helper.

## Why this matters

Stages 3 and 4 of `examples/end_to_end/agent_factory/` use the execution ID to suffix Docker container names (`agent_factory_sandbox_<exec_id>`, `agent_factory_proxy_<exec_id>`, `agent_factory_mock_<exec_id>`) and a per-execution Docker named volume. Without a per-execution suffix, concurrent flow runs collide on a fixed container name and either fail to start or kill an in-use container during teardown.

Any user fork of this pattern (per-flow side resources: containers, tempdirs, named volumes, KV namespaces, etc.) hits the same need. Right now they have to dig into `kitaru.runtime._get_current_execution_id` and hope the underscore-prefixed API doesn't move.

## Proposed shape

Add a public top-level helper:

```python
import kitaru

execution_id = kitaru.current_execution_id()  # str | None
# Returns the active flow's execution ID inside a @kitaru.flow body
# (or @checkpoint), or None outside any flow scope.
```

Or, if `current_*` would be confusing alongside `current_stack`, an alternative:

```python
kitaru.execution_id()           # str | None
kitaru.runtime.execution_id()   # str | None  (namespaced if cluttering top-level is a concern)
```

Either way: stable public re-export of the existing internal accessor, with a docstring that explains "returns None outside a flow run" and a brief example showing per-flow side-resource naming.

## Repro / current usage

```python
# examples/end_to_end/agent_factory/stage_4_credential_proxy.py
from kitaru.runtime import _get_current_execution_id  # private!

@kitaru.flow
def agent_factory_flow() -> str:
    execution_id = _get_current_execution_id() or "local"
    with DockerSandbox(execution_id=execution_id) as sandbox:
        ...
```

Replace with:

```python
import kitaru

@kitaru.flow
def agent_factory_flow() -> str:
    execution_id = kitaru.current_execution_id() or "local"
    with DockerSandbox(execution_id=execution_id) as sandbox:
        ...
```

## Acceptance criteria

- New public helper exported from `kitaru` (or `kitaru.runtime`) returning `str | None`.
- Behaves the same inside a flow / checkpoint as the existing private helper.
- Returns `None` outside any flow scope (matches today's `_get_current_execution_id`).
- Docstring with an example.
- Update the agent_factory stage files to use the public helper.
- Add a unit test or doctest demonstrating in-flow vs out-of-flow behavior.

strickvl:
## What changed

- Added public `kitaru.current_execution_id() -> str | None` as a stable wrapper around the existing runtime execution-ID lookup.
- Exported the helper from top-level `kitaru` and added public API coverage.
- Updated `examples/features/execution_management/wait_and_resume.py` to stop importing the private runtime helper.
- Added a changelog entry.
- Updated `AGENTS.md` and `CLAUDE.md` to say RepoPrompt/orchestration scratch Markdown (plans, reviews, investigations, handoffs, prompt exports) should not be committed unless explicitly requested.

Closes #286.

## Why it was needed

Examples and user code need the current execution ID to name per-run side resources, like Docker containers or named volumes, without reaching into `kitaru.runtime._get_current_execution_id`.

The concrete `agent_factory` usage mentioned in #286 is still in open PR #288, so this PR lands the public helper independently. Once #288 rebases, its stage files can switch to `kitaru.current_execution_id()`.

- Kept `_get_current_execution_id()` unchanged for internal callers.
- Made the public helper a thin delegate so there is still one source of truth for checkpoint-vs-flow priority and out-of-scope `None` behavior.
- Did not add analytics or smoke-test coverage: this is a read-only SDK context accessor, not a new CLI/MCP workflow.
- Removed the planning artifact from the branch and rewrote the PR to keep repo-prompt planning docs out of history.

## Reviewer Notes

Please check the public API shape and the small runtime boundary:

```bash
just test tests/test_kitaru.py tests/test_flow.py tests/test_checkpoint.py
```

I also ran `just check`; format, lint, typecheck, typos, YAML, and actions lint passed, but the offline link step failed on pre-existing links in `prompt-exports/*.md`, unrelated to this PR.

Focus areas:

- `src/kitaru/runtime.py`: public helper delegates directly to the private helper.
- `src/kitaru/__init__.py` and `tests/test_kitaru.py`: export and exact `__all__` update.
- `examples/features/execution_management/wait_and_resume.py`: example now uses the public helper.
- `AGENTS.md` / `CLAUDE.md`: new guardrail against committing RepoPrompt/orchestration scratch Markdown.


htahir1 (dismissed):
cant we directly change the method rather then wrapping it? anyway doesnt matter for now

PR #341, opened 2026-05-17
Update Claude Code skills docs

strickvl:
## What changed

- Updated the public Claude Code skills docs page from a two-skill description to the current three-skill plugin shape.
- Added the quickstart skill, updated the recommended workflow, and refreshed manual install commands to copy all three skill directories.
- Updated authoring coverage and example prompts to mention current adapter families beyond PydanticAI.

## Why it was needed

The public docs lagged behind the `zenml-io/kitaru-skills` plugin. A user following the docs would only see scoping and authoring, missing the quickstart skill and its reference files.

- Kept the page concise and install-focused rather than duplicating the full skill README.
- Manual install now clones the plugin repo and copies full directories, which is important for `kitaru-quickstart` because it has bundled references.
- No native memory guidance was added.

## Reviewer Notes

The important behavior is documentation consistency: the public Kitaru docs should now tell the same story as the plugin README — quickstart first, then scoping, then authoring.

Please check the manual installation snippet carefully. The old curl-based approach copied only individual SKILL.md files; the new approach copies full skill directories so quickstart references are available.

### Reproduction

1. Open `/agent-native/claude-code-skill` in the docs preview.
2. Confirm the page lists quickstart, scoping, and authoring.
3. Confirm the manual install snippet copies all three directories.
4. Confirm example prompts include quickstart and non-PydanticAI adapter prompts.

Local checks run:

```text
git diff --check -- docs/content/docs/agent-native/claude-code-skill.mdx
forbidden native-memory term search on edited page
```

### DIFF
--- a/docs/content/docs/agent-native/claude-code-skill.mdx
+++ b/docs/content/docs/agent-native/claude-code-skill.mdx
@@ -1,18 +1,31 @@
 ---
 title: Claude Code Skills
-description: Install Kitaru's Claude Code skills for scoping and authoring durable agent workflows
+description: Install Kitaru's Claude Code skills for quickstart, scoping, and authoring durable agent workflows
 ---
 
-Kitaru provides two Claude Code skills that teach Claude how to design and build
-durable agent workflows.
+Kitaru provides three Claude Code skills that teach Claude how to discover,
+design, and build durable agent workflows.
 
 ## Available skills
+
+### Quickstart skill
+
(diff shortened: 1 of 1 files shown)


strickvl:
Closing this PR because the cross-repo docs update was opened accidentally during the kitaru-skills work. PR #342 is the focused Kitaru docs cleanup that should continue.

Issue #349, opened 2026-05-19
Confusing Pydantic string_type error when a @checkpoint return value is passed into a Pydantic-validated constructor


htahir1:
## Summary

When a plain `@checkpoint` is called from a flow body on a remote orchestrator, its return value is a `_KitaruOutputArtifact` placeholder. If that placeholder is then passed to a Pydantic-validating constructor (e.g. an adapter's request type), the user sees a Pydantic `string_type` error with the placeholder's repr inside the input value — not a Kitaru-level error explaining the situation.

## Repro

```python
from kitaru import flow, checkpoint
from kitaru.adapters.claude_agent_sdk import (
    ClaudeRunRequest, KitaruClaudeRunner,
)

runner = KitaruClaudeRunner(name="x", options_factory=...)

@checkpoint
def render_prompt() -> str:
    return "Hello, please answer briefly."

@flow
def f() -> str:
    p = render_prompt()
    result = runner.run_sync(
        ClaudeRunRequest.start(p, cwd=".", max_turns=1)
    )
    return result.final_text or "(empty)"
```

Actual error:

```
ValidationError: 1 validation error for ClaudeRunRequest
prompt
  Input should be a valid string [type=string_type,
   input_value=This is a Kitaru checkpoi...o materialize the value.,
   input_type=_KitaruOutputArtifact]
```

## Impact

Composing checkpoint outputs into adapter request objects is a reasonable thing for users to try — especially in a multi-step flow where one checkpoint renders/transforms a value and the next consumes it. The current failure mode mentions `_KitaruOutputArtifact` but doesn't explain why a checkpoint return value can't be used as a literal value, or what the right pattern is.

We hit this while building `examples/integrations/all_adapters_showcase` for a customer call. It blocked a clean "render a prompt, then call the Claude runner" composition.

## Suggested fix (pick one)

- Implement attribute/string coercion on `_KitaruOutputArtifact` so the common composition cases just work, or
- Detect when a placeholder is being passed as a literal value and raise a clear `KitaruUsageError` with: (a) which checkpoint produced the placeholder, (b) where it was consumed, and (c) the right pattern — e.g. "pass placeholder outputs only into other `@checkpoint` calls; or move the consuming logic into a downstream `@checkpoint`."

The error message should surface the Kitaru-level guidance before any Pydantic validation error.

Issue #348, opened 2026-05-19
OpenAI Agents adapter: isinstance(result, OpenAIRunResult) fails inside flow bodies on remote orchestrators


htahir1:
## Summary

When using `KitaruRunner` from inside a `@flow` on a remote stack, an `isinstance` check on the returned result fails even though the value is a fully-populated `OpenAIRunResult`.

## Repro

```python
from kitaru import flow
from kitaru.adapters.openai_agents import (
    KitaruRunner, OpenAIRunRequest, OpenAIRunResult,
)

runner = KitaruRunner(agent, checkpoint_strategy="runner_call")

@flow
def support_flow(question: str) -> str:
    result = runner.run_sync(OpenAIRunRequest.start(question))
    if not isinstance(result, OpenAIRunResult) or result.status != "completed":
        raise RuntimeError(f"OpenAI Agents run did not complete: {result!r}")
    return str(result.final_output)
```

On a remote stack (e.g. a `local_remote` style stack with a remote orchestrator + artifact store), the `isinstance` check evaluates `False`. The exception's `result!r` repr clearly shows `OpenAIRunResult(schema_version=1, status='completed', ...)`, so the object is structurally a complete `OpenAIRunResult` — the class identity has just diverged because `OpenAIRunResult` was re-imported in a different context (the flow body vs. the step that produced the value).

Observed on `develop` with the latest adapter, Python 3.13, ZenML 0.94.x.

## Impact

Defensive patterns users will write naturally (`isinstance` guards on adapter result types) silently misfire and produce confusing "did not complete" errors. The workaround — `getattr(result, "status", None)` — isn't documented.

We hit this while building `examples/integrations/all_adapters_showcase` for a customer call.

## Suggested fix

Either:

- preserve the class identity of public adapter result types across orchestrator import contexts, or
- document the `getattr`/attribute-based pattern as the canonical check in the OpenAI Agents adapter README and update the existing `examples/integrations/openai_agents_agent` to match.

The same concern likely applies to `ClaudeRunResult` and the LangGraph result types — worth a sweep across all adapters.

PR #352, opened 2026-05-19
Make .wait() compose with granular adapter checkpoints, plus all-adapters showcase

### LINKED ISSUE #350: Granular adapter checkpoints become peer terminals — .wait() can't reconcile a clean return value
## Summary

`PydanticAI` granular mode, `OpenAI Agents` `checkpoint_strategy=\"calls\"`, and `LangGraph` `checkpoint_strategy=\"calls\"` all emit one Kitaru checkpoint per model/tool call. When the adapter is invoked from a flow body, these checkpoints have no upstream/downstream relationship to any subsequent flow-body `@checkpoint`, so they all become peer terminals.

Result: `handle.wait()` raises `_MultipleTerminalStepsOutputError`, even when the user writes the obvious "finalize sink" pattern:

```python
@checkpoint
def finalize(result):
    return result.final_output

@flow
def f(q: str) -> str:
    result = runner.run_sync(OpenAIRunRequest.start(q))  # granular: many ckpts
    return finalize(result)  # does NOT become DAG-downstream of runner ckpts
```

Observed terminals (from one of our runs):

```
multiple terminal checkpoints were found (6):
  finalize_pydantic_ai,
  lookup_order_tool,
  pydantic_support_agent_model_request,
  pydantic_support_agent_model_request_2,
  pydantic_support_agent_model_request_3,
  shipping_policy_tool
```

## Root cause

`_extract_outputs_from_terminal_steps` in `src/kitaru/flow.py` walks zenml's `step_spec.upstream_steps`. Adapter-created checkpoints materialize their output back into flow-body scope as a real Python value before `finalize` is called, so zenml's tracer never builds the upstream edge between the adapter's last internal checkpoint and the user's `finalize` step.

## Impact

The supported pattern for granular modes today is `kitaru.save(name, value)` + a post-flow `KitaruClient` artifact lookup (as in `examples/integrations/langgraph_agent`). That works but is significantly heavier than a single `.wait()` call.

More importantly, it undermines the cross-framework pitch: \"every adapter wraps cleanly into a one-line durable flow.\" In a recent prospect conversation (Bloomberg / agent platform team), the questions were exactly about per-tool / per-model checkpointing with clean error reporting back to the agent — which is what granular modes are for. Today, opting into granular mode forces a switch to artifact-lookup-style result extraction, which makes the framework-agnostic story less clean.

We hit this while building `examples/integrations/all_adapters_showcase` (a single demo running the same task through PydanticAI, OpenAI Agents, LangGraph, and Claude Agent SDK). We had to collapse three of the four adapters to invocation-level checkpoint strategies just to keep `.wait()` working.

## Suggested fixes (one or more)

1. **Placeholder-returning adapter API.** Have the adapter return a placeholder rather than a hydrated value when it created Kitaru checkpoints. Passing the placeholder to a downstream `@checkpoint` re-establishes the DAG edge. Inline attribute access (`result.final_output`) would no longer work without an explicit `.materialize()`, which is acceptable for granular mode.
2. **Explicit downstream link helper.** Add a `kitaru.link_downstream(prior_call_id)` or `runner.then(fn)` helper so adapter-internal checkpoints can declare a future sink.
3. **Latest-started-wins heuristic for `.wait()`.** When multiple terminals exist, pick the latest-started one by `started_at` with a warning. Heuristic, but unblocks the common case and matches user intuition (\"the last checkpoint I wrote is the answer\").

(1) is the most architecturally honest; (3) is the cheapest stop-gap.

## Repro

`examples/integrations/all_adapters_showcase/all_adapters.py` on `develop` — run on a remote orchestrator stack. The PydanticAI section in granular mode (no outer `@checkpoint`) reliably produces the 6-terminal error above.

htahir1:
## Summary

Fixes #350 and ships the cross-framework showcase that goes with it.

When a flow body computes a derived value from an adapter's materialized result (e.g. `return str(result.final_output)`) and the adapter created its own granular Kitaru checkpoints, `.wait()` used to raise `_MultipleTerminalStepsOutputError` because each adapter-internal checkpoint appeared as a peer terminal with no DAG sink. After this PR, the user's `return` statement is the authoritative answer for `.wait()` — independent of how many granular checkpoints the adapter created.

## Reviewer Notes

The root cause was structural: `_coerce_flow_return_for_zenml` persists the flow's return value via `save_artifact` from the pipeline-body finalizer, where no step context is active. ZenML's internal `_link_artifact_version_to_the_step_and_model` silently no-ops in that case, so the artifact ends up orphaned — not attached to the run, not in `pipeline_spec.outputs`, not in `step.outputs`. The fallback `_extract_outputs_from_terminal_steps` then saw the granular sibling checkpoints and refused to pick.

The fix records the saved artifact id on the active pipeline run's metadata so the run becomes a discoverable bridge to the orphan artifact. `_get_current_execution_id()` is set by `_flow_scope` and is available throughout the wrapper body, including the post-return finalizer path.

Files to look at:

- `src/kitaru/flow.py`
  - `_save_flow_result_artifact` now also calls `_record_flow_result_artifact_on_run`, which writes `{name: artifact_id}` under a reserved key in the run's metadata. Failures here are non-fatal — the artifact was already saved.
  - `_extract_outputs_from_kitaru_flow_result_artifacts` reads that breadcrumb and rehydrates the artifact via `Client().get_artifact_version`.
  - `_extract_flow_result` inserts the new path *between* the pipeline-spec lookup and the terminal-step fallback, so the existing happy path (placeholder return) is unchanged.
- `tests/test_flow.py` — adds `test_flow_result_extraction_recovers_orphan_kitaru_flow_result_via_run_metadata`. `_DummyRun` gained an opt-in `run_metadata` and `include_output_specs=False` toggle to model the orphan case.

## Reproduction

```bash
uv sync --extra local \
        --extra pydantic-ai \
        --extra openai-agents \
        --extra langgraph-openai \
        --extra claude-agent-sdk
uv run kitaru init
export OPENAI_API_KEY=sk-...
export ANTHROPIC_API_KEY=sk-ant-...
uv run python examples/integrations/all_adapters_showcase/all_adapters.py
```

On a remote stack (e.g. `local_remote`), every adapter completes and the table prints. Before this PR, three of four adapters in their granular modes raised `_MultipleTerminalStepsOutputError`.

Observed locally on `local_remote`:

```
Adapter           Framework         Exec ID   Status     Ckpts  Took
----------------  ----------------  --------  ---------  -----  -----
PydanticAI        pydantic-ai       cfc94bbf  completed  5      43.8s
OpenAI Agents     openai-agents     2e6550a0  completed  5      46.2s
LangGraph         langgraph         a9603542  completed  2      14.3s
Claude Agent SDK  claude-agent-sdk  bdce99c5  completed  1      20.1s
```

Verify in the dashboard or `kitaru executions get <exec_id>` that each granular run has the expected per-model/per-tool checkpoints alongside its single resolved return value.

## Local checks run

- `uv run ruff check src/kitaru/flow.py examples/integrations/all_adapters_showcase/` — clean.
- `uv run pytest tests/test_flow.py -k "flow_result or extract_flow or terminal"` — 88 passed, including the new regression test.
- `uv run pytest tests/ --ignore=tests/mcp` — 1767 passed, 1 unrelated SQLite-lock flake in `tests/test_phase4_compliance_review_stage4.py::test_stage4_flow_waits_resumes_and_reuses_session` (about wait/resume sessions; reproduced before and after this change).

## Other notes

- Bumps a changelog entry under `### Added` (showcase) and `### Fixed` (the underlying fix).
- Does not address #348 (OpenAI Agents `isinstance` class-identity issue on remote) or #349 (placeholder leaks into Pydantic validators). The showcase uses `getattr` rather than `isinstance` for #348 — a documentation note rather than a fix.
- The showcase requires both `OPENAI_API_KEY` and `ANTHROPIC_API_KEY` so it can run against real models. There is no mocked/no-keys mode.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,15 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+- Added an all-adapters showcase (`examples/integrations/all_adapters_showcase/`) that runs the same customer-support scenario through PydanticAI, OpenAI Agents, LangGraph, and Claude Agent SDK and prints one consolidated comparison table — demonstrating per-adapter checkpoint granularity (typically 5/5/2/1) with one durable-flow boundary on a remote stack.
+
 ### Changed
 - Bumped the minimum ZenML dependency, server image, and Helm subchart versions to `0.94.4` so Kitaru tracks the latest upstream ZenML release.
 
 ### Fixed
 - Fixed Kitaru flow return compatibility with ZenML `0.94.4` dynamic-pipeline output validation by persisting plain flow returns as internal artifacts while preserving user-facing Python return values and avoiding marker-shaped user dictionaries being mistaken for hidden tuple metadata.
+- Fixed `.wait()` raising `_MultipleTerminalStepsOutputError` when a flow body returns a derived value alongside granular adapter checkpoints (PydanticAI granular, OpenAI Agents `checkpoint_strategy="calls"`, LangGraph `checkpoint_strategy="calls"`). The flow-return coercer now records the saved artifact id on the active pipeline run's metadata, and `_extract_flow_result` consults that breadcrumb before falling back to terminal-step extraction. Fixes #350.
 
 ## [0.12.0] - 2026-05-17
 
--- a/examples/integrations/all_adapters_showcase/README.md
+++ b/examples/integrations/all_adapters_showcase/README.md
@@ -1,0 +1,90 @@
+# All adapters showcase
+
+One customer-support scenario, four agent frameworks, one Kitaru durable-execution
+layer. The point of this example is to show that the `@flow` / `@checkpoint`
+boundary is identical no matter which agent harness a team picks — only the
+inner agent changes.
+
+## What it runs
+
+A single business task: *"Where is order ORD-1007? Look it up, apply the
+shipping policy, tell me what happens next."*
+
+The same task is executed through every adapter Kitaru ships today:
+
+| # | Adapter | Module |
+|---|---|---|
+| 1 | PydanticAI | `kitaru.adapters.pydantic_ai.KitaruAgent` |
+| 2 | OpenAI Agents SDK | `kitaru.adapters.openai_agents.KitaruRunner` |
+| 3 | LangGraph (calls mode) | `kitaru.adapters.langgraph.KitaruGraphRunner` |
+| 4 | Claude Agent SDK | `kitaru.adapters.claude_agent_sdk.KitaruClaudeRunner` |
+
+Each adapter is wrapped in the same `@flow` shape:
+
+```python
+@flow
+def some_framework_flow(question: str) -> str:
+    return runner.run_sync(question).output
+```
+
+That's the value: a single primitive (`@flow`) gives every adapter the same
+durable execution, checkpoint tracking, artifact persistence, and replay
+surface — without locking the team into any framework.
+
+## Running it
+
+```bash
+uv sync --extra local \
+        --extra pydantic-ai \
+        --extra openai-agents \
+        --extra langgraph-openai \
+        --extra claude-agent-sdk
+uv run kitaru init
+
+export OPENAI_API_KEY=sk-...
+export ANTHROPIC_API_KEY=sk-ant-...
+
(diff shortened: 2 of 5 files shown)


strickvl:
Addressed by #355 already

PR #373, opened 2026-05-22
Add Replay Lab model matrix kit

strickvl:
## What changed

This adds a generic Replay Lab model-matrix kit on top of the base Replay Lab branch.

The change includes:
- plural candidate support in Replay Lab reports
- a core post-replay evaluator hook
- a trusted-local gate for MCP evaluator descriptors
- canonical replay-trust and candidate-verdict policy in the core report
- aggregate Replay Lab analytics events
- a live LangGraph requirements-triage model-matrix example
- a shared verdict HTML renderer that presents core report decisions
- sanitized sample report JSON/HTML artifacts
- a public-safe `MODEL_MATRIX_KIT.md` walkthrough
- a folder-specific `langgraph_requirements_triage/README.md` for the live example
- a generic coordinating Replay Lab skill
- smoke-test coverage for deterministic Replay Lab plus help-only checks for the live path

## Why it was needed

The base Replay Lab demo proves the replay/comparison idea with a deterministic support flow. This stacked change adds the model-matrix story: replay the same cohort against multiple candidate model aliases, score the outputs after replay, and produce a report that helps decide whether a cheaper/faster/better candidate is safe enough to inspect or ship.

The docs and example now start from the production problem rather than the solution: a requirements-triage agent is already making small but important decisions, and the team wants evidence before changing the model or prompt behind it.

- Replay Lab internals are split into focused modules for models, validation, evaluator loading, verdict policy, reporting, and small utilities. `kitaru._replay_lab` remains the compatibility facade.
- The report schema is plural-first: `candidates`, `candidate_results`, and `effect_vs_baseline`.
- Candidate IDs are explicit and required.
- `candidate_descriptor` remains an input convenience alias, but reports are plural-only.
- Evaluator metadata lives under `metrics.evaluation`; evaluator scores can override quality by default without changing runtime facts like cost or latency.
- MCP evaluator descriptors are rejected by default because they can import and execute local Python code. A local server owner can opt in explicitly with `KITARU_MCP_ALLOW_LOCAL_EVALUATORS=1`.
- Replay trust is canonical per case and in the summary. The older `replay_drift_warning` field is kept as a derived/legacy compatibility field.
- Candidate verdicts require replay trust, quality evidence, and efficiency evidence before a candidate can be marked `ship`.
- The HTML renderer reads canonical verdict/trust/decision-evidence data emitted by the core report instead of recomputing a separate truth.
- The live example uses `current` for observed seeding and keeps `cheap`, `balanced`, and `quality` as replacement candidates, so the report does not make the current model look like a proposed winner.
- The live example uses generic `requirements_triage` naming and alias-only model inputs; no partner names or provider model strings are committed.

## Reviewer Notes

The main behavior now flows through focused Replay Lab modules under `src/kitaru/`. The short version: Replay Lab starts from observed executions, creates a baseline replay, fans out candidate replays, optionally evaluates all completed lanes, then writes one canonical report model. Think of it as widening the original three-lane road into a model-matrix road: the observed and baseline lanes are the guardrails, and each candidate gets its own comparison result against the same baseline.

The highest-risk part to inspect is the trust/verdict path. Earlier review flagged that the JSON/Markdown and HTML renderer could accidentally develop separate definitions of “high replay drift” or “ship/caution/hold.” This PR now makes the core report emit canonical replay trust, per-candidate aggregate verdicts, decision evidence ordering, and overall recommendation. The HTML renderer should mostly present those fields, not invent a second verdict system.

The MCP evaluator descriptor gate is also worth reviewing. Local SDK/script usage can still load trusted evaluator callables, but MCP clients cannot pass local Python evaluator descriptors unless the server owner starts the MCP server with `KITARU_MCP_ALLOW_LOCAL_EVALUATORS=1`. In concrete terms: a local script can still say “use this evaluator function,” but a casual MCP request cannot quietly cause the server to import and run local Python code.

The live LangGraph example is intentionally opt-in. The flow builds the graph/model inside the flow body from a Kitaru model alias so replay can actually swap models. If that graph/model ever moves back to module import time, the model matrix can silently become fake because every candidate would use the same already-bound model.

### Reproduction

Keyless deterministic path:

```bash
uv run kitaru init
uv run python examples/end_to_end/replay_lab/seed_observed.py --small
uv run python examples/end_to_end/replay_lab/run_replay_lab.py
uv run python examples/end_to_end/replay_lab/render_report.py
```

Then open:

```text
examples/end_to_end/replay_lab/reports/support-replay-lab-demo.html
```

Model-matrix sample report, no live model calls:

```bash
uv run python examples/end_to_end/replay_lab/langgraph_requirements_triage/render_report.py \
  --json-path examples/end_to_end/replay_lab/langgraph_requirements_triage/reports/requirements-triage-sample.json \
  --output-path /tmp/requirements-triage-sample.html
```

Then open `/tmp/requirements-triage-sample.html` and check that it shows candidate verdicts, replay trust, candidate decision evidence, changed-output sections, and evaluator details.

Optional live path requires registered Kitaru model aliases. Use `current` for the observed lane and candidate aliases such as `cheap`, `balanced`, and `quality` for replacements:

```bash
uv run kitaru model register current --model <provider/model-for-current-alias>
uv run kitaru model register cheap --model <provider/model-for-cheap-alias>
uv run kitaru model register balanced --model <provider/model-for-balanced-alias>
uv run kitaru model register quality --model <provider/model-for-quality-alias>

uv run python examples/end_to_end/replay_lab/langgraph_requirements_triage/seed_observed.py --small --model current
uv run python examples/end_to_end/replay_lab/langgraph_requirements_triage/run_replay_lab.py --candidate-limit 2
uv run python examples/end_to_end/replay_lab/langgraph_requirements_triage/render_report.py
```

### Local checks run

```bash
uv run pytest tests/test_replay_lab.py tests/test_replay_lab_example.py tests/mcp/test_server.py
just check
just test
```

Results:
- simplify-pass focused Replay Lab/MCP tests: `152 passed`
- full suite: `1933 passed, 8 skipped`
- `just check`: passed


strickvl:
Shipped in #469. This was just a prototype.

PR #389, opened 2026-06-03
Add adapters docs section

strickvl:
## What changed

- Added a first-class `Adapters` section to the docs navigation.
- Moved the five framework adapter pages out of `Guides` and into shorter canonical `/adapters/.../` slugs.
- Added an Adapters overview page that explains the adapter mental model and helps readers choose the right page.
- Added permanent redirects from the old `/docs/guides/*-adapter` URLs to the new adapter pages.
- Updated hand-written docs and example README links to point at the new canonical adapter URLs.

## Why it was needed

The adapter docs had become important entry points for users who arrive with an existing agent framework, but they were buried at the bottom of the generic Guides section. This change gives those pages their own home while preserving old links.

- The new canonical slugs drop the repeated `-adapter` suffix because the section name already supplies that context: `/docs/adapters/pydantic-ai/`, `/docs/adapters/openai-agents/`, etc.
- The old guide URLs stay valid through `RETIRED_DOCS_REDIRECTS`, so existing bookmarks and old docs links do not 404.
- The moved adapter pages are pure file moves; their content is preserved except for canonical links elsewhere.

## Reviewer Notes

The important behavior happens in two places.

First, FumaDocs now sees `docs/content/docs/adapters/` as a normal top-level section because `docs/content/docs/meta.json` lists `adapters` after `guides`, and `docs/content/docs/adapters/meta.json` defines the section order. If this is wrong, the docs build might still succeed, but the sidebar would either omit the new section or place the adapter pages awkwardly.

Second, old public URLs are handled in `docs/worker/docs-routing.mjs`. A request for `/docs/guides/pydantic-ai-adapter/` should hit the retired-docs redirect before the Worker tries to serve static assets. If this is wrong, old links from search results, changelogs, or external READMEs could break.

The main thing to inspect manually is the new overview page at `docs/content/docs/adapters/index.mdx`: it should feel like a practical entry point, not a marketing page, and it should avoid promising more replay precision than the adapters actually provide.

### Reproduction

1. Build and validate the docs export:

   ```bash
   just generate-docs
   just docs-build
   just docs-validate
   ```

2. In the built docs, confirm the sidebar has a top-level **Adapters** section after **Guides**.

3. Open `/docs/adapters/` and confirm it links to:
   - `/docs/adapters/pydantic-ai/`
   - `/docs/adapters/openai-agents/`
   - `/docs/adapters/claude-agent-sdk/`
   - `/docs/adapters/gemini-interactions/`
   - `/docs/adapters/langgraph/`

4. Confirm an old URL such as `/docs/guides/pydantic-ai-adapter/` redirects to `/docs/adapters/pydantic-ai/`.

Local checks run:

- `just generate-docs`
- `just docs-build`
- `just docs-validate`
- stale-link search confirmed the only remaining old guide adapter URLs are intentional redirect sources in `docs/worker/docs-routing.mjs`.


strickvl:
Addressed the Markdown redirect gap: old `/docs/guides/*-adapter.md` paths now redirect to the new `/docs/adapters/*.md` canonical Markdown pages. Re-ran `just docs-build` and `just docs-validate` after rebasing; validation passes.

strickvl:
Addressed the Markdown link rewriting gap too: added `adapters` to the docs-root rewrite whitelist and taught the materializer to rewrite MDX `href="/..."` attributes, not only Markdown `[text](/...)` links. Verified after `just docs-build` that materialized Markdown has no bare `/adapters/...` links/hrefs, then reran `just docs-validate` successfully.

Issue #130, opened 2026-04-11
Post-merge hardening for clean/info commands


strickvl:
## Context

PR #65 (clean/info enhancements) has been thoroughly reviewed across 6 independent reviews. The PR is solid and mergeable, but several hardening items were identified that are worth addressing as follow-ups.

### Defense-in-depth (low effort, high value)

- [ ] **Add repo root guard to `_validate_deletion_target`** — If `KITARU_CONFIG_PATH` is misconfigured to point at the repo root, the current guards (/, $HOME, cwd) wouldn't catch it when running from a subdirectory. The plan already has `repo_root` resolved; add one more check. (`_cleanup.py:587`)

- [ ] **Add `_validate_deletion_target` call before `shutil.rmtree` in `_reinitialize_project`** — Currently bypasses the safety guard. Low-risk in practice (target is always `<project>/.kitaru/`) but good hygiene. (`_cleanup.py:638`)

- [ ] **Add `exc_info=True` to two log statements** — Graceful shutdown failure log (`_local_server.py:373`) and `_describe_local_server_for_cleanup` second exception handler (`_cleanup.py:428-432`). Makes failures diagnosable.

### Backup robustness

- [ ] **Copy WAL/SHM sidecar files alongside SQLite backup** — Under WAL mode, committed data can live in `-wal` until checkpoint. After a force-kill, the backup may be incomplete without sidecars. (`_cleanup.py:551-560`)

- [ ] **Add microseconds to backup timestamp** — `strftime("%Y-%m-%dT%H-%M-%S")` has second-level resolution. Add `-%f` to avoid theoretical collision in CI/automation. (`_cleanup.py:415`)

### Preview accuracy

- [ ] **Fix `zen_server` size undercounting** — Preview only counts `zen_server/daemon/` but the catch-all loop skips the entire `zen_server` directory. Other content (logs, PID files) is invisible in the total. (`_cleanup.py:258-271`)

- [ ] **Handle `size_bytes=0` vs `None` in `_render_preview_entry`** — Currently `if entry.size_bytes` is falsy for 0, showing no size string for empty dirs. Use `is not None` if 0 B should be shown. (`_cli/_clean.py:50`)

### Test coverage

- [ ] **Test model registry protection with `alias_count=None`** — The "unreadable config" path where `--force` is still required has no dedicated test. High-value safety test.

[...]

PR #394, opened 2026-06-04
Show zero-byte sizes in cleanup preview

GauravG-20:
## What changed?

Updated cleanup preview rendering to distinguish between `size_bytes=0` and `size_bytes=None`.

Previously, `_render_preview_entry()` used a truthiness check when deciding whether to render a size string. As a result, valid zero-byte entries were displayed without a size in dry-run output.

This change replaces the truthiness check with an explicit `is not None` check and adds a regression test covering both empty files and empty directories.

This addresses the "Handle size_bytes=0 vs None in _render_preview_entry" item from Issue #130.

## Why?

A size of `0` is a valid value and should not be treated the same as an unknown size (`None`).

Before this change, zero-byte files and directories appeared in cleanup previews without any size information, making the output inconsistent with other preview entries.

## Reviewer Notes

The behavior change is isolated to preview rendering in `_render_preview_entry()`.

The previous implementation used:

```python
if entry.size_bytes
```

which treated `0` and `None` identically. The new implementation explicitly checks for `None`, allowing valid zero-byte sizes to be rendered as `0 B`.

The regression test exercises the behavior through the existing `clean project --dry-run` workflow rather than testing the rendering helper directly. This validates the end-to-end preview generation and rendering path.

## Reproduction

1. Create a Kitaru project containing an empty file and an empty directory inside `.kitaru/`.
2. Run:

```bash
kitaru clean project --dry-run
```

Before this change, the empty file and directory were displayed without a size.

After this change, both entries are rendered with:

```text
(0 B)
```

## Local checks run

* `just test tests/test_cli.py::TestCleanProject::test_dry_run_shows_zero_byte_entry_sizes`
* Verified that the new regression test fails with the previous implementation and passes with the fix


strickvl (approved):
Thanks for this contribution!

PR #427, opened 2026-06-11
Fix chatbot image PydanticAI deps

### LINKED ISSUE #424: examples/chatbot: unpinned 'pydantic-ai' in ImageSettings breaks deployed pods (adapter import error)
## Bug

`examples/chatbot/chatbot.py` declares:

```python
CHATBOT_IMAGE = ImageSettings(
    requirements=["pydantic-ai", "openai"],
    ...
)
```

The unpinned `pydantic-ai` resolves to the latest release at image build time. Recent pydantic-ai releases (≥1.104) are incompatible with the Kitaru adapter, so every fresh deploy of the example builds an image whose pods die **before the first checkpoint**:

```
File ".../exercises/.../chatbot.py", line 38, in <module>
  from kitaru.adapters.pydantic_ai import KitaruAgent, wait_for_input
File ".../kitaru/adapters/pydantic_ai/__init__.py", line 20, in <module>
  from ._agent import KitaruAgent
File ".../kitaru/adapters/pydantic_ai/_agent.py", line 32, in <module>
  from pydantic_ai.agent.abstract import (...)
# ImportError: cannot import name 'AgentInstructions' from 'pydantic_ai.agent.abstract'
```

From the invoking client's perspective the execution just shows `failed` with no checkpoints and a truncated failure message — very confusing to debug (took a live traceback dump to find). A UI invoking the deployment looks "broken/glitchy" because every session dies ~30s in.

**Verified:** kitaru 0.15.0; the same incompatibility reproduces locally with `pip install kitaru pydantic-ai` (latest) — `pip install 'kitaru[pydantic-ai]'` (pins `pydantic-ai-slim>=1.89,<1.104`) fixes both local and in-pod imports.

## Fix

One line in the example:

```python
requirements=["kitaru[pydantic-ai]", "openai"],
```

Consider the same pattern for other examples that bake framework deps into `ImageSettings`, and/or surface a clearer error when an adapter import fails inside a pod (the `failure.message` was literally `)`).

strickvl:
## What changed

Fixes #424.

The chatbot example image now installs Kitaru with its supported PydanticAI and OpenAI extras:

```python
requirements=["kitaru[pydantic-ai,openai]"]
```

instead of installing bare `pydantic-ai` and `openai`. This keeps the deployed image under Kitaru's adapter-tested dependency bounds and prevents resolver backtracking to incompatible PydanticAI versions.

I also added regression coverage for Kitaru extras in image requirements, updated the chatbot README local setup command to install both needed extras, and updated the GitBook docs so they no longer teach bare `pydantic-ai` image requirements.

## Why it was needed

The chatbot image previously resolved to `pydantic-ai`, `openai`, and auto-added plain `kitaru`. Plain `kitaru` does not install `kitaru[pydantic-ai]`, so the deployment image could choose a PydanticAI package outside Kitaru's supported adapter range. The reported failure happens before checkpoints because `examples/chatbot/chatbot.py` imports `kitaru.adapters.pydantic_ai` at module import time.

The local README setup also needed to include the OpenAI extra because the chatbot model is `openai:gpt-4o-mini`.

- Reused the existing image requirement conversion logic. `_requirements_include_kitaru()` already treats `kitaru[...]` extras as Kitaru requirements, so no production config code needed to change.
- Added regression tests for both `kitaru[pydantic-ai]` and `kitaru[pydantic-ai,openai]` to ensure plain `kitaru` is not appended again.
- Kept `news_scout` dependency compatibility and improved PydanticAI import diagnostics out of this PR; they are related follow-ups with separate dependency/adapter decisions.

## Reviewer Notes

The important behavior now happens in the example image declaration and the existing image conversion path. The chatbot asks for `kitaru[pydantic-ai,openai]`; when Kitaru converts that `ImageSettings` into Docker requirements, it sees the requirement already names Kitaru and leaves it alone. If this is wrong, a built chatbot image would either fall back to bare `pydantic-ai` again or add a duplicate plain `kitaru` requirement.

The main files to inspect are:

- `examples/chatbot/chatbot.py` for the fixed image requirement.
- `examples/chatbot/README.md` for matching local setup instructions.
- `tests/test_config.py` for the no-duplicate Kitaru-extra regression coverage.
- `tests/test_chatbot_example_contract.py` for the source-level chatbot contract test. It intentionally uses AST inspection so the test does not need to import the optional-dependency-heavy chatbot module just to check constants.
- `docs/book/guides/containerization.md` for the docs explanation that auto-injection adds plain `kitaru`, not optional extras.

### Reproduction

To see the fixed behavior end to end without building a remote image, run:

```bash
uv run python - <<'PY'
from examples.chatbot.chatbot import CHATBOT_IMAGE
from kitaru.config import image_settings_to_docker_settings
print(image_settings_to_docker_settings(CHATBOT_IMAGE).requirements)
PY
```

Expected output:

```text
['kitaru[pydantic-ai,openai]']
```

That proves the chatbot image now keeps the Kitaru extras requirement and does not auto-append a second plain `kitaru`.

Local checks run:

```bash
uv run pytest tests/test_config.py tests/test_chatbot_example_contract.py
just check
just test
```

Latest focused checks after the README update:

- `uv run pytest tests/test_config.py tests/test_chatbot_example_contract.py`: `215 passed, 20 warnings`
- `just check`: passed

Earlier full-suite result on this branch: `just test` → `2385 passed, 8 skipped, 86 warnings`.

### DIFF
--- a/docs/book/getting-started/quickstart.md
+++ b/docs/book/getting-started/quickstart.md
@@ -157,13 +157,18 @@ parameter:
     stack="prod-k8s",
     image={
         "base_image": "python:3.12-slim",
-        "requirements": ["httpx", "pydantic-ai"],
+        "requirements": ["kitaru[pydantic-ai,openai]", "httpx"],
         "apt_packages": ["git"],
     },
 )
 def research_agent(topic: str) -> str:
     ...
 ```
+
+This example includes `kitaru[pydantic-ai,openai]` explicitly because setting
+`base_image` means you control the image contents. Kitaru can auto-add plain
+`kitaru` when it builds the requirements list for you, but it does not guess
+optional extras such as the PydanticAI/OpenAI adapter dependencies.
 
 See the [Containerization guide](../guides/containerization.md) for the full set of
 image options, custom Dockerfiles, and how Kitaru packages your source code.
--- a/docs/book/guides/containerization.md
+++ b/docs/book/guides/containerization.md
@@ -31,7 +31,7 @@ import kitaru
 @flow(
     image=kitaru.ImageSettings(
         base_image="python:3.12-slim",
-        requirements=["httpx", "pydantic-ai"],
+        requirements=["kitaru[pydantic-ai,openai]", "httpx"],
         apt_packages=["git"],
         environment={"MY_VAR": "value"},
     ),
@@ -46,7 +46,7 @@ Or as a dictionary:
 @flow(
     image={
         "base_image": "python:3.12-slim",
-        "requirements": ["httpx", "pydantic-ai"],
+        "requirements": ["kitaru[pydantic-ai,openai]", "httpx"],
     },
 )
 def my_agent(topic: str) -> str:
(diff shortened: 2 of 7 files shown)


strickvl:
Locally validated ✅

PR #431, opened 2026-06-11
Fix streaming wait tools

### LINKED ISSUE #425: PydanticAI adapter: event_stream_handler is incompatible with flow-scope wait tools (allow_sync_tool_body_waits) — blocks streaming durable chatbots
## Problem

The two flagship adapter features cannot be combined today:

- **Durable chat pattern** (`examples/chatbot`): a tool body must run at **flow scope** so it can call `wait_for_input()` (waits can't be created inside checkpoints). This works via `tool_checkpoint_config_by_name={"say_and_wait": False}` + `allow_sync_tool_body_waits=True`, whose mechanism is keeping sync tool bodies **on the workflow thread**.
- **Live streaming**: token-level live events (`pydantic_ai.stream.*`) require an `event_stream_handler`, which switches the run to PydanticAI's streamed path.

In the streamed path, sync tool bodies are dispatched via `run_in_executor` **inside the model-request processing** — i.e., inside the model checkpoint scope (contextvars propagate to the executor thread). The tool body then fails immediately:

```
File "pydantic_ai/_utils.py", line 111, in run_in_executor
File "chatbot.py", in say_and_wait
  persist_history(list(conv.history))
File "kitaru/checkpoint.py", line 419, in _assert_call_allowed
kitaru.errors.KitaruContextError: Nested checkpoint calls are not supported in the Kitaru MVP.
```

Even without the history checkpoint, `wait_for_input()` would fail next (wait created inside checkpoint scope).

**Net effect:** you can have a durable chatbot OR a streaming agent, not a streaming durable chatbot. This is the first thing a user building a chat UI will try (it was the first thing we tried). Verified on kitaru 0.15.0.

## Proposal: deferred tool execution in streaming mode

When a tool is checkpoint-opted-out (and/or `allow_sync_tool_body_waits` is set), the adapter could:

1. Stream the model request to completion inside its checkpoint (collecting the tool call + emitting live events as today),
2. **exit the model checkpoint**, execute the opted-out tool at flow scope on the workflow thread (waits + explicit checkpoints like `persist_history` work as in the non-streaming path),
3. start the next streamed model request.

PydanticAI's `iter` API already exposes the run node-by-node (model request nodes vs tool call nodes), so the adapter can schedule tools between checkpoints rather than within them.

## Bonus observation for chat UIs

In the `say_and_wait` pattern the user-visible message travels as a **tool argument**, so even with streaming enabled it arrives as `part_delta` events with `delta_kind=tool_call` (partial JSON of the `message` arg) rather than text deltas. Worth either (a) documenting how a UI can incrementally render tool-arg deltas, or (b) considering a first-class "assistant message" streaming surface for the chat pattern once deferred tool execution lands.

Related: the chatbot example README could note this limitation until fixed (currently a user adding `event_stream_handler` to the example gets the nested-checkpoint error with no hint why).

strickvl:
Fixes #425

## What changed

This PR fixes the PydanticAI adapter path where `event_stream_handler` forced a whole-turn adapter checkpoint and left explicitly opted-out sync tool bodies running inside that checkpoint.

The change adds a private streaming-fallback marker for adapter-created turn checkpoints, then allows only an explicitly opted-out ordinary sync `FunctionToolsetTool` to run with checkpoint scope temporarily suspended. This keeps the existing public rules intact: `checkpoint()` still rejects real nested checkpoints, and `wait()` still requires flow scope.

It also makes `run_stream(...)` and `iter(...)` enter the same inline sync-tool execution context as `run(...)` / `run_sync(...)` when `allow_sync_tool_body_waits=True`.

## Why it was needed

The durable chatbot pattern uses a sync tool like `say_and_wait` to persist history and then wait for user input. The tool opts out of Kitaru's synthetic tool checkpoint with `tool_checkpoint_config_by_name={"say_and_wait": False}` and opts into workflow-thread sync execution with `allow_sync_tool_body_waits=True`.

Without this fix, adding `event_stream_handler` caused Kitaru to open one large adapter turn checkpoint around the upstream PydanticAI run. The tool skipped only its own synthetic checkpoint; it still ran inside the larger adapter checkpoint. Then `persist_history(...)` failed as a nested checkpoint, and `wait_for_input(...)` would fail next because waits cannot be created inside checkpoint scope.

- The escape is private to the PydanticAI adapter; no core checkpoint or wait guard was weakened.
- The suspension is narrow: active streaming-fallback marker, `allow_sync_tool_body_waits=True`, explicit per-tool `False`, currently inside checkpoint scope, and `FunctionToolsetTool.is_async is False`.
- Generic toolsets, async tools, and `@hitl_tool` remain out of scope for this PR.
- The private marker is set only for calls-strategy runs that are forced into whole-turn checkpointing by streaming/event-handler behavior.

## Reviewer Notes

The important behavior starts in `KitaruAgent._run_sync(...)` / `_run_async(...)`. When a calls-strategy run is forced into whole-turn checkpointing because of streaming hooks, Kitaru marks that adapter-owned checkpoint. Later, when PydanticAI calls a tool, `KitaruToolset.call_tool(...)` checks whether this exact tool is an explicitly opted-out sync `FunctionToolsetTool`. If so, it briefly clears Kitaru checkpoint scope and ZenML `StepContext` while that one tool body runs.

The thing to inspect carefully is the guard around that suspension. If it is too broad, Kitaru could accidentally make arbitrary checkpoint-contained waits legal. If it is too narrow, the chatbot pattern still fails. The new unit tests in `tests/test_pydantic_ai_adapter.py` cover the positive path and the negative cases: no marker, default config, dict config, `allow_sync_tool_body_waits=False`, async function tools, generic tool objects, HITL tools, and user-created checkpoint contexts.

The full issue reproduction is in the new phase17 integration test. It runs an event-handler agent turn, calls an explicit `@kitaru.checkpoint(cache=False)` helper from the opted-out sync tool, then reaches the real ZenML wait.

### Reproduction

Run the focused regression test:

```bash
uv run pytest -q tests/test_phase17_pydantic_ai_adapter.py::test_phase17_direct_wait_tool_with_event_stream_handler_reaches_wait
```

Expected result: the child process reports that it reached the real ZenML wait instead of failing with nested checkpoint / checkpoint-contained wait errors.

For the narrow guard behavior, run:

```bash
uv run pytest -q tests/test_pydantic_ai_adapter.py -k 'streaming_fallback or inline_sync_tool_context'
```

Expected result: the eligible sync `FunctionToolsetTool` sees flow scope, while generic, async, HITL, and non-opted-out tools do not get the suspension.

### Local checks run

```bash
just check
just test
```

`just test` result: `2393 passed, 8 skipped, 97 warnings in 90.79s`.

### DIFF
--- a/src/kitaru/adapters/pydantic_ai/_agent.py
+++ b/src/kitaru/adapters/pydantic_ai/_agent.py
@@ -60,7 +60,9 @@ from ._utils import (
     CheckpointConfig,
     CheckpointStrategy,
     ToolCheckpointOverrides,
+    adapter_streaming_fallback_checkpoint,
     checkpoint_input_value,
+    has_any_explicit_tool_checkpoint_opt_out,
     run_async_in_checkpoint,
     run_sync_in_checkpoint,
     turn_cache_key,
@@ -95,6 +97,7 @@ class _TurnCheckpointCallConfig:
     checkpoint_inputs: dict[str, Any]
     checkpoint_config: CheckpointConfig
     force_turn_checkpoint: bool
+    mark_streaming_fallback_checkpoint: bool
 
 
 # Auto-flow bodies keyed by uuid. The @kitaru.flow entrypoint must be module-
@@ -109,13 +112,6 @@ _AUTO_FLOW_LOCK = threading.Lock()
 
 if f"src.{__name__}" not in sys.modules:
     sys.modules[f"src.{__name__}"] = sys.modules[__name__]
-
-
-def _has_tool_checkpoint_opt_out(
-    overrides: ToolCheckpointOverrides | None,
-) -> bool:
-    """Return whether any named tool explicitly opts out of checkpointing."""
-    return bool(overrides and any(override is False for override in overrides.values()))
 
 
 def _strategy_from_granular_checkpoints(
@@ -474,7 +470,7 @@ class KitaruAgent(WrapperAgent[AgentDepsT, OutputDataT]):
                 "`granular_checkpoints=True`, and a matching checkpoint opt-out "
                 'such as `tool_checkpoint_config_by_name={"tool_name": False}`.'
             )
-        if allow_sync_tool_body_waits and not _has_tool_checkpoint_opt_out(
+        if allow_sync_tool_body_waits and not has_any_explicit_tool_checkpoint_opt_out(
             tool_checkpoint_config_by_name
         ):
             raise KitaruUsageError(
@@ -794,6 +790,9 @@ class KitaruAgent(WrapperAgent[AgentDepsT, OutputDataT]):
                 disable_cache=force_turn_checkpoint,
             ),
             force_turn_checkpoint=force_turn_checkpoint,
+            mark_streaming_fallback_checkpoint=(
+                force_turn_checkpoint and self._uses_calls_strategy
+            ),
         )
 
     async def _auto_checkpoint_async(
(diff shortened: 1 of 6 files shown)


strickvl:
Locally validated ✅

PR #451, opened 2026-06-16
Make prospect scout example genuinely agentic

htahir1:
## What this does

Reworks the `examples/end_to_end/prospect_scout/` example (shipped in 0.16.0) so it is a **genuine agent**, not a workflow wearing an agent costume.

The first version called `search_company_signals()` directly and pasted the results into the prompt — the model never decided anything, and `KitaruAgent` was reduced to a typed-output wrapper. By the prospect's own definition (see below), that's a workflow, not an agent.

Now `research_prospect` hands the qualifier a `search_web` **tool** and a company name, and the agent decides which searches to run (hiring, funding, expansion, layoffs) and how many. Because it runs through `KitaruAgent`, every `search_web` call is tracked as a child event under that company's checkpoint — so the agent's research is observable, not a black box.

## Why

This example demonstrates a realistic **agentic** prospecting workflow (the kind a staffing or B2B sales team runs): heavy on web search, type-safe classification into lines of business, durable execution ("if it dies after 47 of 60, don't redo the 47"), human-in-the-loop, and cost control. The original version demonstrated none of the *agentic* part — the model never chose what to search. The useful distinction: a workflow doesn't need to contemplate tool choice; an agent does. "Make the model contemplate tool choice" is the bar, and that's what this change delivers.

## Key changes

- **`search_web` is a real PydanticAI tool** the agent chooses to call; topic-aware fixtures (per company, per topic) make multiple searches return genuinely different results offline, so the agent's tool-choice behavior is meaningful without an Exa key.
- **`LineOfBusiness` enum** added to the typed `ProspectAssessment` (alongside `FitLevel`) — the staffing-desk classification the use case needs, enforced by validation + PydanticAI retry.
- **Agents built lazily inside their checkpoints** (`new_qualifier()` / `new_outreach_writer()`), so module import needs no key and remote pods can inject keys via a secret before the checkpoint runs. (Building at module scope crashed at import — I hit exactly this during retry testing.)
- **`OutreachDraft` typed output** carrying its company, and `research_prospect` now overrides `assessment.company` with the company it researched. Both keep every persisted artifact distinct, which fixes a latent `UNIQUE constraint failed: step_run_input_artifact` flake when identical model outputs (e.g. under `TestModel`) fan into a sink.
- **README reorganized around four "aha moments"**: agent-chooses-its-searches, crash→retry caching/cost, type-safety, and HITL-without-idling.
- **Regression test** `test_qualifier_actually_calls_search_web` asserts the agent invokes its tool — the check that would have caught the original miss.

## Reviewer Notes

The story lives entirely in `prospector.py`. The risky parts:

1. **The agent loop** (`research_prospect` + `new_qualifier` + `search_web`). If `search_web` weren't actually registered/called, we'd be right back to a workflow — hence the regression test. `search_web` calls `kitaru.log`, which only works inside a flow; that's fine in the checkpoint, and the test stubs it.
2. **Artifact distinctness** (`OutreachDraft.company`, the `assessment.company` override). Remove either and the fan-out into `publish_report` / the list into `build_shortlist` can collide on identical `TestModel` artifacts. I reproduced the `IntegrityError`, applied the fix, and then ran the happy path repeatedly with no recurrence.
3. **Lazy agent construction.** `publish_report(shortlist_handle, drafts)` still receives the shortlist handle to keep `build_shortlist` connected to the single sink (avoiding `_MultipleTerminalStepsOutputError`).

### Reproduction

Keyless, deterministic, full happy path including the HITL gate:

```bash
PROSPECT_SCOUT_MODEL=test python examples/end_to_end/prospect_scout/prospector.py \
    --companies "Apex BioLabs,Helios Energy,Northwind Logistics"
# when it pauses, in another shell:
PROSPECT_SCOUT_MODEL=test kitaru executions input <id> --value true
```

Confirm the agent actually searched:

```bash
uv run pytest tests/test_prospect_scout_example.py::test_qualifier_actually_calls_search_web -q
```

Durability / cost "aha moment" — crash mid-sweep, then retry (note: the retry shell needs the same model env, since it re-imports the module):

```bash
PROSPECT_SCOUT_MODEL=test PROSPECT_SCOUT_CRASH_AFTER=1 python .../prospector.py --companies "Apex BioLabs,Helios Energy"
PROSPECT_SCOUT_MODEL=test kitaru executions retry <id>
kitaru executions get <id>   # research_prospect (cached) — not re-searched
```

### Local checks run

`ruff check` / `ruff format` / `ty check` / `uvx typos` / `scripts/audit-example-coverage.py` all pass; `tests/test_prospect_scout_example.py` is 11/11. I ran the happy path (1–3 companies) and a clean crash→retry against the staging server end to end; the cached-checkpoint behavior and the approval gate both work.


htahir1:
@strickvl — context + a request to take this one home, like you did with #446.

**Background.** This reworks the `prospect_scout` example from #446. After it shipped it turned out the example wasn't doing the thing it claimed: `search_company_signals()` was a fixed function call pasted into the prompt, so the qualifier never *chose* to search — and `KitaruAgent` was just a typed-output wrapper. The useful distinction: a workflow doesn't need to contemplate tool choice; an agent does. This was a workflow in disguise.

**What changed (full detail in the PR description):**
- `search_web` is now a real tool the agent decides to call; topic-aware fixtures make multi-search meaningful offline.
- Added a `LineOfBusiness` enum to the typed output (the staffing-desk classification).
- Agents are built lazily inside their checkpoints so remote pods can inject keys via a secret (module-scope construction crashed at import — I hit this during retry testing).
- `OutreachDraft` carries its company + `research_prospect` overrides `assessment.company`, which fixes a latent `IntegrityError: UNIQUE constraint failed: step_run_input_artifact` flake when identical `TestModel` outputs fan into a sink. Repro'd it, fixed it, then ran the happy path repeatedly clean.
- README reorganized around four "aha moments"; added a regression test asserting the agent actually invokes its tool.

**What I verified against staging — all with `TestModel`/fixtures (no provider key in my env):** keyless happy path (1–3 companies) through the HITL gate, and a clean crash → `kitaru executions retry` showing `research_prospect` cached (not re-searched). All local checks + the 11 example tests pass.

**What I could NOT verify and would love you to pressure-test with a real key:**
1. The **crash → retry** durability path with a real model + real search key. One gotcha I documented: the retry shell needs the same model env because the retry re-imports the module.
2. **Type-safety retry** — `TestModel` always emits a valid enum, so I never triggered the validation-retry path; worth forcing a bad classification.
3. Real **incurred-vs-reused token numbers** to back the cost claim (under `TestModel` cost is $0, and the retried run reported `LLM usage: none`).
4. Whether the `OutreachDraft` / `assessment.company` artifact-distinctness fix is the right place to solve that integrity collision, or whether it points at something to harden in ZenML's list-input handling.

Happy to iterate on anything. Thanks!

strickvl:
@htahir1 sure thing. I'll fix the audit stuff as well when I check it over.

htahir1:
@strickvl actually i did most of it myself, including deploying it remotely, and fixing CI audits and other CI bugs. if u can approve ill merge

strickvl (approved):
Approved assuming you tested it on local + cloud stack as you stated in the comments.

PR #472, opened 2026-06-25
Add `fast-agent` adapter experimental preview

strickvl:
## What changed

This PR ships an experimental-preview fast-agent adapter for Kitaru.

It adds `KitaruFastAgent`, a calls-mode adapter that enters a normal fast-agent run context, waits for the yielded `AgentApp`, discovers reachable agents, and wraps supported model/tool methods so they become Kitaru checkpoints while the surrounding Kitaru flow is running.

The preview supports reachable:

- LLM `generate(...)`
- LLM `structured(...)`
- LLM `structured_schema(...)` when the underlying LLM exposes it
- `agent.call_tool(...)`
- supported calls on detached clones returned by `spawn_detached_instance(...)`

It also adds:

- `kitaru[fast-agent]` optional extra, pinned to a narrow known-good `fast-agent-mcp` range
- OpenAI-backed fast-agent integration example using `gpt-5-nano`, with a no-network memory fallback for deterministic tests
- GitBook adapter docs and index/TOC/example wiring
- example coverage metadata and release smoke coverage
- analytics event for adapter wrapping
- deterministic tests and real-runtime smoke coverage
- best-effort `llm_usage_v1` token usage records for fast-agent model-call checkpoints
- `save_usage` and `cost_calculator` support on `KitaruFastAgent`
- OpenAI Responses provider aliases so fast-agent's `responses` provider can be priced by `genai-prices`

## Why it was needed

Raw fast-agent users get the fast-agent runtime behavior, but not Kitaru replay checkpoints around internal model/tool calls. This adapter lets a user keep fast-agent in charge of agent behavior while Kitaru records the calls it can actually see and safely replay.

The extra usage/cost work matters because otherwise a fast-agent run could be durable but financially opaque: the model-call checkpoint would exist, but Kitaru would not show how many tokens were spent or the likely model cost.

This is deliberately marked as an experimental preview because the current implementation depends on live fast-agent runtime object shapes rather than an official upstream hook API.

- The first shipped strategy is `checkpoint_strategy="calls"` only.
- The adapter fails loudly if the yielded app has no discoverable agent mapping, rather than pretending durability is active when nothing was wrapped.
- Optional LLM methods are only exposed by the wrapper when the original LLM actually has them. For example, the wrapper does not make `structured_schema(...)` appear on a generate-only LLM.
- `runtime="isolated"` is rejected for adapter-created checkpoints because the synthetic checkpoint body closes over live fast-agent model/tool objects.
- Usage capture happens inside the synthetic model-call checkpoint body, after the original model method returns and before the checkpoint body returns.
- Usage extraction checks all reachable fast-agent usage accumulators first, then falls back to known result fields such as `usage` / `usage_metadata`.
- Usage capture is best-effort: malformed usage metadata never turns a successful provider call into a failed checkpoint.
- Empty usage records are skipped. The checkpoint already proves the call happened; `llm_usage_v1` is only logged when token counts are available.
- The public example now uses real fast-agent `ToolAgent` / `AgentApp` objects with OpenAI `gpt-5-nano` by default, so reviewers can inspect checkpoint-level `llm_usage_v1` token usage and `genai-prices` estimated cost. A `--provider memory` path remains for deterministic no-network tests and local smoke checks.
- The optional extra is currently constrained to `Python >=3.13.5,<3.14` and `fast-agent-mcp>=0.4.48,<0.4.49` because newer fast-agent releases currently conflict with Kitaru/ZenML dependency pins.

Remaining fast-agent adapter gaps are tracked in #473.

## Reviewer Notes

The important behavior lives in `src/kitaru/adapters/fast_agent/_wrapping.py`, `src/kitaru/adapters/fast_agent/_usage.py`, and `src/kitaru/adapters/fast_agent/_utils.py`.

The concrete chain is:

1. User creates their fast-agent app normally.
2. User wraps it with `KitaruFastAgent(fast)`.
3. `KitaruFastAgent.run()` enters `fast.run()`.
4. After fast-agent yields an `AgentApp`, Kitaru finds app agents from `_agents`, `agents`, or `active_agents`.
5. Kitaru replaces reachable model/tool methods with small wrappers.
6. When fast-agent later calls the LLM or calls a tool through `agent.call_tool(...)`, Kitaru opens a synthetic checkpoint around that exact call.
7. On model calls, Kitaru snapshots reachable usage accumulators before the provider call, then logs `llm_usage_v1` inside the same checkpoint body if token counts appear.
8. On replay, if the logical inputs match, Kitaru can reuse the checkpoint result instead of calling the model/tool again.

The highest-risk review areas are:

- method wrapping and idempotency in `_wrapping.py`
- usage extraction and best-effort failure behavior in `_usage.py`
- stable checkpoint identity in `_utils.py`
- optional-extra dependency markers and uv conflict declarations in `pyproject.toml`
- docs honesty in `docs/book/adapters/fast-agent.md`
- the OpenAI-backed example and `--provider memory` fallback in `examples/integrations/fast_agent_agent/fast_agent_adapter.py`

### Reproduction

Run the normal project checks:

```bash
just check
just test
just example-coverage-audit
```

Then verify the preview optional extra and example path:

```bash
uv run --python 3.13.5 --extra fast-agent --no-dev pytest \
  tests/test_fast_agent_calls_strategy.py \
  tests/test_fast_agent_adapter_foundation.py \
  tests/test_fast_agent_example.py \
  tests/test_fast_agent_usage.py
```

Run the no-network deterministic fallback end to end:

```bash
tmpdir=$(mktemp -d)
ZENML_CONFIG_PATH="$tmpdir/zenml" \
ZENML_REPOSITORY_PATH="$PWD" \
ZENML_ANALYTICS_OPT_IN=false \
uv run --python 3.13.5 --extra fast-agent --no-dev \
  python examples/integrations/fast_agent_agent/fast_agent_adapter.py --provider memory
```

Expected result: the script prints a fast-agent adapter demo summary and reports a submitted Kitaru execution. In the output, reviewers should see checkpoints named like:

- `fast_agent_demo_generate_model_call`
- `fast_agent_demo_uppercase_tool_call`

The model-call checkpoints should include `llm_usage_v1` metadata with provider `memory`, model `memory-fast-agent-demo`, and deterministic local word-count usage. This path intentionally has no estimated dollar cost because the fake memory model is not priceable.

Live OpenAI check run locally with `OPENAI_API_KEY`:

```bash
uv run python examples/integrations/fast_agent_agent/fast_agent_adapter.py
uv run kitaru executions get <execution-id> --output json
```

Latest successful execution checked: `5944d972-c1fb-4647-a70e-37f0ce8790ab`. Its three model-call checkpoints recorded:

- `provider: responses`
- `model: gpt-5-nano`
- `cost.source_label: genai-prices`
- estimated checkpoint costs: `0.00002815`, `0.00003415`, and `0.0000127` USD

Note: this calls-mode example deliberately does not call `handle.wait()` because adapter-created model/tool checkpoints are multiple terminal steps. Inspect checkpoint-level `llm_usage_v1` records on the execution instead.

Local checks run before push:

- `/simplify` pass with reuse / quality / efficiency review agents; applied cleanup findings
- Oracle review loop; fixed the reported blockers
- `just check` ✅
- `just test` ✅ `2950 passed, 16 skipped`
- `just example-coverage-audit` ✅
- fast-agent extra/no-dev targeted tests ✅ `34 passed`
- live OpenAI `gpt-5-nano` example verification ✅ (`5944d972-c1fb-4647-a70e-37f0ce8790ab`)
- fast-agent OpenAI example/docs follow-up checks ✅


strickvl:
Decided against adding this in so as not to spread ourselves too thin.

PR #494, opened 2026-07-02
Use unique terminal result candidate

strickvl:
## What changed

This PR tightens `FlowHandle` result extraction for replay/adapter runs where terminal-step inference sees several terminal checkpoints, but only one terminal checkpoint is still eligible to represent the flow result.

When multiple terminal steps exist, Kitaru now checks the existing `flow_result_candidate` marker before raising ambiguity. If exactly one terminal step remains eligible, `.wait()` / `.get()` can return that checkpoint output. Adapter-created model/tool checkpoints that are explicitly marked `flow_result_candidate=False` no longer force ambiguity by themselves.

This PR also fixes result-source precedence for saved flow returns:

- linked `kitaru_flow_result_ref_v1` artifacts are loaded before terminal-step heuristics;
- explicit `None` flow returns are recorded with `kitaru_flow_result_none_v1` and also beat terminal-step heuristics;
- explicit run outputs and declared pipeline output specs still win before either metadata marker.

Finally, `lychee.toml` now excludes `CHANGELOG.md` from link checks. The changelog preserves historical release links, so a stale old docs URL should not block current PR CI.

## Why it was needed

PR #477 fixed the plain saved-flow-return path, but a nearby replay shape could still fail:

1. A PydanticAI granular run creates terminal model/tool checkpoints.
2. The user flow also has a final explicit result checkpoint, e.g. `publish_support_decision`.
3. The adapter-created checkpoints are marked as non-result candidates.
4. Kitaru still saw several terminal steps and raised `_MultipleTerminalStepsOutputError` instead of returning the one remaining user-result checkpoint.

That meant the execution could complete successfully, but `.wait()` failed while trying to read the result.

The review also found the opposite bad outcome for newer runs with linked flow-result metadata:

1. The flow returns a plain value.
2. Kitaru saves that value as the `kitaru_flow_result` artifact and records its artifact ID under `kitaru_flow_result_ref_v1`.
3. The same run also has terminal adapter/tool checkpoints.
4. Terminal inference could return a heuristic checkpoint before Kitaru read the explicit linked result.

That meant `.wait()` / `.get()` could return a checkpoint value instead of the actual flow return value.

There was one more edge case for explicit `None`: if a flow intentionally returned `None`, there was no saved result artifact to link. Kitaru now records a lightweight metadata marker for that case. In a real run, if that marker cannot be written, Kitaru fails closed rather than later guessing a terminal checkpoint result. In direct/local wrapper calls with no execution id, it skips the marker because there is no execution to read later.

- Preserved explicit source precedence: run-level outputs first, then declared pipeline output specs.
- Moved linked `kitaru_flow_result_ref_v1` recovery before terminal-step inference.
- Added explicit `None` result metadata via `kitaru_flow_result_none_v1` before terminal-step inference.
- Kept terminal-step inference as the legacy fallback when no linked result or explicit-None metadata exists.
- Preserved ambiguity when zero candidates or multiple eligible candidates remain.
- Factored single-terminal output extraction into a helper so the one-terminal and one-eligible-terminal paths share the same validation.
- Excluded `CHANGELOG.md` and internal planning/review scratch folders from lychee link checks.

## Reviewer Notes

The important behavior is in `src/kitaru/flow.py` inside `_extract_flow_result`, `_load_referenced_flow_result`, and `_extract_outputs_from_terminal_steps`.

The result-source order is now:

1. explicit `run.outputs`
2. declared `snapshot.pipeline_spec.outputs`
3. linked `kitaru_flow_result_ref_v1`
4. explicit `kitaru_flow_result_none_v1`
5. terminal-step inference

So for a newer run, if Kitaru recorded “the actual flow result is artifact `abc123`,” `.wait()` reads `abc123` before it guesses from graph terminals. If the actual flow result was `None`, `.wait()` returns `None` before it guesses from graph terminals. For an older run without either marker, Kitaru still falls back to the terminal-step logic.

The terminal-step behavior is still intentionally narrow. When terminal steps are `adapter_call`, `adapter_tool_call`, and `finalize`, Kitaru asks: “after excluding steps marked `flow_result_candidate=False`, is there exactly one possible result step?” If yes, it loads that step’s single output. If no, it keeps raising the ambiguity error.

The risks to inspect are:

- linked result metadata should not override explicit run outputs or declared output specs;
- explicit `None` metadata should not override explicit run outputs, declared output specs, or linked result metadata;
- linked result and explicit `None` metadata should override terminal-step guesses;
- old/no-marker runs should keep using terminal fallback, including the existing error cases;
- `CHANGELOG.md` should stay out of lychee checks, while current docs links remain covered.

### Reproduction

Run the focused regressions:

```bash
uv run pytest tests/test_flow.py -k "linked_flow_result or linked_result or unique_terminal_result_candidate or terminal_filter or zero_output_terminal or none_marker or explicit_none" -n 0
```

The key linked-result and `None` tests are:

- `test_linked_flow_result_wins_over_unique_terminal_candidate`
- `test_linked_flow_result_wins_over_single_non_candidate_terminal`
- `test_linked_flow_result_wins_over_zero_output_terminal`
- `test_broken_linked_flow_result_does_not_fall_back_to_terminal_output`
- `test_flow_result_none_marker_wins_over_unique_terminal_candidate`
- `test_flow_result_none_marker_wins_over_zero_output_terminal`
- `test_flow_wrapper_records_explicit_none_with_discovered_execution_id`

For the CI link-check fix:

```bash
just links-external
```

Local checks run:

```bash
uv run pytest tests/test_flow.py::test_flow_return_coercion_records_explicit_none_result tests/test_flow.py::test_flow_return_coercion_fails_closed_when_none_marker_fails tests/test_flow.py::test_flow_runtime_scope_keeps_execution_id_none_without_zenml_context tests/test_flow.py::test_flow_wrapper_records_explicit_none_with_discovered_execution_id tests/test_flow.py::test_flow_result_none_marker_wins_over_unique_terminal_candidate tests/test_flow.py::test_flow_result_none_marker_wins_over_zero_output_terminal tests/test_flow.py::test_linked_flow_result_ref_beats_none_marker tests/test_flow.py::test_flow_result_run_outputs_beat_none_marker tests/test_flow.py::test_flow_result_output_specs_beat_none_marker tests/test_flow.py::test_flow_result_without_none_marker_uses_terminal_fallback -n 0
just links-external
just check
just test
```

Latest results:

- focused explicit-`None`/linked precedence subset: `10 passed`
- `just links-external`: `711 Total`, `681 OK`, `0 Errors`, `30 Excluded`
- `just check`: passed
- full suite: `3219 passed, 27 skipped`


strickvl:
Addressed the linked-result precedence review finding in the latest push. Result extraction now uses explicit run outputs, then declared output specs, then linked `kitaru_flow_result_ref_v1`, and only then terminal-step heuristics. Added regressions for linked result vs unique terminal candidate, single non-candidate terminal, zero-output terminal, and broken linked metadata not falling back to terminal output.

strickvl:
Latest push handles the extra review edge cases and the lychee failure:

- preserved linked `kitaru_flow_result_ref_v1` precedence before terminal inference;
- added explicit `None` flow-result metadata via `kitaru_flow_result_none_v1`, including fail-closed behavior when a real execution cannot record the marker;
- kept direct/local wrapper calls with no execution id returning `None` normally;
- added regressions for `None` beating terminal heuristics, including zero-output terminal fallback;
- excluded `CHANGELOG.md` from lychee checks so historical release links do not fail PR CI.

Checks run locally:

- focused explicit-`None`/linked precedence subset: `10 passed`
- `just links-external`: `711 Total`, `681 OK`, `0 Errors`, `30 Excluded`
- `just check`: passed
- `just test`: `3219 passed, 27 skipped`

PR #587, opened 2026-07-17
Rewrite README around trace-and-replay positioning

htahir1:
## What and why

The README led with a four-layer taxonomy and framed Kitaru as "durable execution for agents." This rewrite repositions the GitHub landing page around the new story: **production is the eval** — every agent run is captured as a replayable **trace**, replay re-executes your real code against the recorded world, and forking one variable at a time turns production traffic into a regression suite.

Key changes:

- **Hero**: one-paragraph pitch (capture → replay → fork → trust the diff), no scaffolding prose. Deployment is presented as optional — Kitaru records underneath whatever framework runs the agent, and *can* also run/deploy it.
- **Why Kitaru?**: opens with the transcript-vs-re-execution contrast and the "debugger with a memory, sitting beside your observability stack" line.
- **The loop** (replaces the old Quick Start + Python-first sections): one running example — a zero-config PydanticAI agent wrapped in `KitaruAgent` (no decorators, no `.wait()`), then replay from the agent's first model call as the baseline, fork by patching the recorded `refund_payment_tool` output, widen into a tagged regression sweep over recent traces.
- **Durable execution** demoted to a single "plumbing" paragraph.
- **Adapters table** now lists all six shipped adapters with verified export names.
- **New "Drive it from your coding agent" section**: CLI + MCP setup (`claude mcp add kitaru -- kitaru-mcp`) and the kitaru-skills plugin install commands.
- 313 → 225 lines. Scarf pixel, asset paths, and existing links preserved.

## Reviewer Notes

The risky part of this PR is **accuracy of the code story**, not prose. Everything in "The loop" was checked against the source rather than written from memory:

- Zero-config wrapping and per-call checkpoints: `docs/book/adapters/pydantic-ai.md` ("No flow decorator, no checkpoint annotations… auto-opens one").
- Synthetic checkpoint names: tool checkpoints as `<tool>_tool` (adapter docs replay-semantics section) and model requests as `{agent_name}_model_request` (`src/kitaru/adapters/pydantic_ai/_model.py:598`) — if either naming scheme changes, the README example silently breaks.
- Replay/fork/widen calls: `client.executions.replay(...)` signature (`at=`, `checkpoint_overrides=`, list input, `tag=`) and `executions.list/latest` verified in `src/kitaru/client.py`.

[...]

PR #591, opened 2026-07-19
Complete case-first replay journey

strickvl:
## Summary

This PR completes a case-first replay loop that starts with a production trace and ends with a durable, scored regression result. It connects four previously separate concerns into one path:

1. Register an Agent and immutable AgentVersion.
2. Import Langfuse observations as attributed, inspectable executions.
3. Evaluate stored evidence or run a registered candidate against it.
4. Persist the attempt, membership, lineage, scores, limits, evidence quality, and verdict so it can be inspected or rerun.

The source import remains immutable throughout. Candidate runs create new executions linked to the imported original rather than mutating or re-executing the historical record.

### Agent identity and lifecycle

Registered Agents now have canonical SDK, CLI, and MCP lifecycle surfaces. Registration records a stable Agent identity, an immutable implementation version, source/configuration fingerprints, and an executable entrypoint without running or deploying the Agent. Labels cannot silently move between versions, repeated unchanged registration is idempotent, and direct PydanticAI execution rejects missing registration or identity drift.

The new `kitaru agents ...`, `KitaruClient.agents`, top-level SDK exports, and six Agent MCP tools share the same inspection contract. The older Project-named management APIs remain as hidden compatibility delegates for one release; connection configuration still deliberately uses Project terminology.

### Attributed Langfuse imports

`KitaruClient.imports.langfuse(...)` and `kitaru import langfuse` now accept either an observations JSONL export or a read-only `langfuse://trace/<id>` fetch. Each import declares exactly one Agent and AgentVersion. Kitaru verifies supported Langfuse version evidence against that declaration, distinguishes provider-verified from caller-attributed traces, and rejects conflicts before writing anything.

Dry-run is the default. A write requires both `--write` and `--confirm-data-storage`. Successful imports preserve the selected raw rows and a normalized, versioned replay-evidence bundle as separate immutable artifacts, together with hashes, source attribution, graph integrity, bounded diagnostics, historical usage/cost, and capability-specific replay readiness. Existing legacy imports remain readable without being guessed into a version or rewritten.

### Scoring stored evidence

The new scoring API evaluates one or many stored executions without rerunning the Agent. Scorers declare whether they are pure, grounded against explicitly allowlisted read-only capabilities, or comparative against an execution's immediate original. Every invocation appends a typed `SCORED`, `ABSTAINED`, `BLOCKED`, or `ERROR` observation.

Scorer source/configuration snapshots, immutable descriptive aggregates, score history, latest-valid resolution, and bounded score filtering make the evidence reproducible and queryable. Replay attempts can attach the same scorers to verified candidate children after execution.

### Durable experiments and regression verdicts

Replay and score submissions now persist as experiments with frozen ordered membership, candidate attribution, repeat policy, checkpoint coverage, planning rows, verified child membership, idempotency, and paginated read models. Execution projections expose immediate original, root lineage, derived replays, and related experiments. Verified imported replay children now expose their logical parent and root through the same public execution projection while native replay lineage remains authoritative.

Registered-Agent replay can rerun a frozen suite with explicit trial, token, cost, and duration limits. Replay verdicts combine objective scores, pinned protections, operational limits, and evidence completeness into `PASS`, `FAIL`, or `HOLD`. Missing or non-comparable evidence cannot be converted into an optimistic zero or an accidental pass. When a suite name matches multiple attempts, the CLI now returns its normal concise text or JSON usage error and directs the caller to an exact experiment ID instead of leaking a traceback.

### Imported PydanticAI candidate replay

A registered PydanticAI `KitaruAgent` can start a candidate from either the recorded root input or a complete recorded model-message/tool-result boundary. Kitaru reconstructs the validated prefix, checks candidate/source tool compatibility, serves exact recorded tool responses without invoking live read or write tools, and blocks every miss.

Each child records immutable response-match decisions and comparability separately from its objective/protection scores. Root-input reruns are explicitly counterfactual and default to `HOLD`; a complete boundary can reach `PASS` only when the candidate stays on the comparable recorded path and every score and protection passes.

### Public walkthrough and documentation

The replay example now tells the case-first journey with deterministic Langfuse fixtures and a registered PydanticAI support Agent. The walkthrough initializes the example directory itself as the Kitaru repository while isolating state through `KITARU_CONFIG_PATH`, registers the named source and fixed candidates through direct `agent.register(...)` SDK calls, and imports evidence only through the public `kitaru import langfuse` command.

The story starts with one account-setting case, reproduces its recorded path, labels the weakened root-input run as an expected counterfactual `HOLD`, then widens to a second case. It finishes with a bounded comparable `PASS`, an idempotent rerun, exact experiment-ID inspection, parseable read-only JSON, and public parent/root lineage on the candidate child. Ambiguous suite-name inspection now fails cleanly and tells the reader to use the exact experiment ID.

The docs add or expand the Agent lifecycle, Langfuse import, scoring, PydanticAI imported replay, and replay-safety guides. Smoke tests and example coverage metadata exercise the public commands and keep the example contract visible.

## Measured scope

The complete diff against `feat/langfuse-trace-import` contains **165 changed files, 42,385 additions, and 4,216 deletions**.

| Area | Files | Additions | Share of additions | Deletions |
|---|---:|---:|---:|---:|
| Library source (`src/`) | 57 | 21,820 | 51.5% | 421 |
| Tests (`tests/`) | 60 | 17,378 | 41.0% | 1,305 |
| Examples | 25 | 1,878 | 4.4% | 1,557 |
| Hand-written docs, README, and changelog | 15 | 1,098 | 2.6% | 447 |
| Configuration, scripts, and CI support | 8 | 211 | 0.5% | 486 |

Within `src/`, the largest additions are imports (4,521 lines), scoring (4,291), PydanticAI adapter support (3,890), and experiments (3,884). Tests account for 41% of all additions and cover registration, attribution and import atomicity, evidence normalization/loading, scoring, experiment planning/execution, verdicts, PydanticAI recorded-tool behavior, and the end-to-end journey.

The branch also contains a merge of newer `develop` changes that were not present on the PR base. That accounts for the smaller cleanup, dependency, CI, and compatibility changes outside the replay feature itself.

## Current limitations

- **Imported candidate replay is implemented only for the PydanticAI adapter.** LangGraph, Google ADK, Claude Agent SDK, and other adapters cannot yet execute candidates from imported Langfuse evidence.
- **Imported executions are immutable historical records.** They cannot be resumed, retried, cancelled, or passed through native checkpoint replay. Imported candidate replay always creates separate child executions.
- **Candidate replay is an SDK operation.** The CLI imports and inspects the resulting durable objects, but it does not launch imported PydanticAI candidate replay.
- **Only complete validated boundaries are eligible.** Message-history replay accepts complete recorded model messages or tool results. Root-input reruns are counterfactual; incomplete, divergent, or blocked paths result in `HOLD` or non-comparable evidence.
- **Recorded tools have no live fallback.** Responses are served only on an exact recorded match; changed arguments, order, occurrence, schema, or implementation are blocked, including for read-only tools.
- **Langfuse support is currently inbound.** This PR imports observations from JSONL or a trace URI; it does not write Kitaru scores, verdicts, or replay traces back to Langfuse.
- **No dashboard implementation is included.** The new contracts are exposed through the Python SDK, CLI, MCP tools, stored artifacts, and the example walkthrough.
- **Imported payloads are not automatically redacted.** Raw observations may contain prompts, customer data, tool arguments/results, or secrets, which is why writes require explicit storage consent.
- **Standalone scoring aggregates are descriptive.** `client.executions.evaluate(...)` records scores and aggregates but does not itself issue regression verdicts or enforce a CI gate. `PASS`/`FAIL`/`HOLD` belongs to protected replay experiments.

## Reviewer Notes

The most important behavior to review is the transition from untrusted historical observations to executable candidate evidence. Import must bind every trace to an exact registered version without guessing, persist raw and normalized evidence atomically and idempotently, and never execute source code during preview or import.

For candidate replay, focus on the recorded-response boundary. A matching PydanticAI tool call must receive the exact stored response without invoking the callable. Any mismatch or contract drift must be blocked and reflected in comparability so it cannot produce an unsupported `PASS`.

Finally, inspect the experiment and scoring persistence paths together. Membership, scorer revisions, aggregate inputs, protections, limits, candidate attribution, and verdict evidence must remain immutable and resolvable after the process that submitted the attempt exits.

### Reproduction

The complete one-case-first walkthrough is in `examples/end_to_end/replay_fork_demo/README.md`. Start from the main checkout, but initialize the example directory itself as the Kitaru repository so its real entrypoints stay inside the repository boundary:

```bash
uv sync --extra local --extra pydantic-ai --extra llm --extra langfuse
cd examples/end_to_end/replay_fork_demo

if [ -e .kitaru ]; then
  echo "Stop: replay_fork_demo already has a .kitaru marker." >&2
  return 1 2>/dev/null || exit 1
fi

export KITARU_REPLAY_DEMO_DIR="$PWD"
export KITARU_REPLAY_CONFIG_DIR="$(mktemp -d /tmp/kitaru-replay-demo.XXXXXX)"
export KITARU_CONFIG_PATH="$KITARU_REPLAY_CONFIG_DIR"
uv run kitaru init
```

`KITARU_CONFIG_PATH` gives the walkthrough an isolated database, settings, and artifact store. Keep using the same shell. Set `OPENAI_API_KEY` before importing `evals.register`; registration constructs the configured objects but does not call the model.

Register the recorded baseline through the public SDK, not through `demo.py`:

```bash
uv run python - <<'PY'
from evals.register import baseline_agent
from reference_agent.config import IMPORTED_SOURCE_VERSION

baseline_agent.register(
    label=IMPORTED_SOURCE_VERSION,
    entrypoint="evals.register:baseline_agent",
)
PY

uv run kitaru agents show support-agent
```

Preview one selected case through the public import command, then repeat it with `--write --confirm-data-storage` and copy the resulting execution ID into `ACCOUNT_ID`:

```bash
uv run kitaru import langfuse \
  trace_fixtures/imported-support-cases.jsonl \
  --source-project-id langfuse-replay-example \
  --agent support-agent \
  --agent-version v2.3-structured-escalation-imported \
  --trace-id support-account-setting
```

The first candidate command continues from the validated recorded tool-result boundary and can produce the comparable baseline reproduction:

```bash
uv run python demo.py resume "$ACCOUNT_ID" \
  --name account-setting-reproduction
```

Check for `PASS`, `message_history from tool-result`, `recorded_path_comparable`, one recorded reply served, zero misses or blocked calls, no path divergence, and passing objective/protection rows. A live provider can still diverge, in which case the honest result is `HOLD` or `FAIL`.

The weakened candidate deliberately starts from the root input. State the expected verdict before running it:

```bash
uv run python demo.py replay "$ACCOUNT_ID" \
  --name account-setting-counterfactual
```

This is expected to be `HOLD` because root-input evidence is counterfactual even when its objective and protection pass. The README next registers `mini_tool_budget_2_agent` directly with `label="permissions-fix-v1"`, tests that candidate from the same validated boundary, and only then imports `support-service-status`. The resulting ordered two-case root-input experiment is also an explicit `HOLD`, not a manufactured multi-case pass.

The final comparable check reruns the frozen single-case request with explicit bounds:

```bash
uv run python demo.py rerun account-setting-fix \
  --candidate-variant mini_tool_budget_2 \
  --candidate-version permissions-fix-v1 \
  --idempotency-key account-setting-fix-rerun-v1 \
  --max-trials 1 \
  --max-cost-usd 0.10 \
  --max-incurred-tokens 100000 \
  --max-duration-seconds 300
```

The intended final result is a comparable `PASS`; `HOLD` or `FAIL` exits nonzero. Run the exact command again to confirm the explicit idempotency key returns the stored attempt without another provider call or duplicate spend.

After the rerun, use the exact experiment ID rather than the now-ambiguous suite name:

```bash
uv run kitaru agents experiments support-agent
export FINAL_EXPERIMENT_ID=<experiment-id-from-the-rerun>
uv run kitaru agents experiments support-agent "$FINAL_EXPERIMENT_ID"

export FINAL_CHILD_ID=<candidate-child-execution-id>
uv run kitaru executions get "$FINAL_CHILD_ID"
```

An ambiguous suite selector now exits through the normal concise CLI error boundary in text and JSON modes. Exact IDs remain resolvable. The child execution's public projection shows the imported execution as both its immediate parent and replay root.

### Validation

Final checks after the simplify pass:

- `just check` passed, including format, lint, type checking, typos, YAML, Actions lint, and offline links.
- `just test tests/test_replay_fork_demo.py tests/test_replay_fork_acceptance.py` passed: **23 passed**.
- Focused CLI ambiguity/error-boundary tests passed: **4 passed**.
- Focused public imported-lineage tests passed: **6 passed**.
- The integration pass over the full non-live suite completed with **4,195 passed, 29 skipped**, plus one unrelated concurrent import-writer race failure that passed when rerun serially. A second full run later stalled and was interrupted. This is not being represented as a completely green final full-suite run.
- A bounded live OpenAI rehearsal confirmed direct source/candidate registration, source-verified import, the comparable message-history `PASS`, the weakened root-input `HOLD`, the fixed-candidate `PASS`, and the two-case root-input `HOLD`. The final one-trial comparable rerun stayed within its explicit bounds, returned `PASS`, and an exact repeat reused the stored experiment without another provider call. Exact-ID inspection and read-only JSON parsing succeeded, the public child projection showed its imported parent/root lineage, and no original tool callable ran during imported replay.

No credentials, raw prompts, tool payloads, or machine-local paths are included in these notes.


htahir1:
Ran the full `replay_fork_demo` journey today from a clean checkout of this branch (isolated worktree + fresh `KITARU_CONFIG_PATH`, live OpenAI calls with the fixture traces). The machinery is genuinely impressive end to end — register → preview → source-verified import → boundary inspection → imported replay with recorded tool serving → immutable experiments with verdict reason codes all worked exactly as documented. Two findings from the run-through, both in the example rather than the SDK: as committed, the example's happy path doesn't complete — every candidate run ends `failed → hold`.

## 1. The output validator swallows the validation error (verified fix)

All three of my candidate runs (`baseline`/gpt-5-mini ×2, `mini_tool_budget_2`/gpt-5-nano ×1) failed identically: the model returns `SupportDecision` JSON with invented values for the `Literal` fields — e.g. `risk_status="blocked - requires human review"`, `policy_label="admin_actions_requires_human_review"` — then blows the single output retry and the run fails with `UnexpectedModelBehavior`.

Root cause is a one-two punch in `reference_agent/agent.py`:

- `_shared_instructions()` names the JSON fields but never the **allowed values** for `policy_label` / `risk_status` / `required_action`, so the model guesses; and
- the retry hides the correction signal:

```python
raise ModelRetry("Return one valid SupportDecision JSON object.") from exc
```

Feeding the error back fixes it — I verified this with a one-line change:

```python
raise ModelRetry(
    "Return one valid SupportDecision JSON object. "
    f"The previous output failed validation: {exc}"
) from exc
```

Same command, same boundary, same variant: the run went from `failed` to **`completed` with objective mean 1.0 and the `completed-execution` protection passing**. I'd also suggest enumerating the literal values in `_shared_instructions()` and/or giving the agent `output_retries=2` for headroom — three consecutive failures across two models suggests provider drift since the fixtures were authored, and the example shouldn't be one schema-guess away from a FAIL.

## 2. Free-text args on `escalate_to_human` make a clean reproduction ~impossible

Even the completed run stays `hold`: the candidate escalates (correct behavior) but phrases the free-text escalation arguments differently from the recording, so exact-match yields `argument_mismatch` → the recorded write response is blocked → `recorded_response_hits: 0` → `imported_replay_not_comparable`. Fail-closed blocking on writes is the right call, but with a live model and free-text args, README step 4's "clean reproduction reports one recorded-response hit" is effectively unreachable.

Two possible shapes for a fix, in ascending ambition:

- **Example-level:** make `escalate_to_human`'s signature structured and minimal (e.g. `customer_id` + `policy_label`) so a well-behaved candidate can actually match the recording.
- **Adapter-level (can wait):** let the tool's `kitaru_replay` metadata declare which arguments are identity-relevant for recorded-response matching (`match_args: ["customer_id"]`), so cosmetic free-text differences degrade evidence quality instead of hard-blocking. That would also generalize to the `search_kb`-rephrasing case the docs already call out.

Happy to share the full sandbox transcript (execution IDs, experiment inspections, logs) if useful. To be clear about what went *right*: attribution came back `source_verified`, idempotency-key reuse returned the stored attempt rather than re-running, the blocked write produced a typed decision with `block_reason`, and the verdict engine refused to invent a pass at every step — the honest-by-default behavior is very convincing once the two issues above are fixed.

strickvl:
@htahir1 I addressed both issues from [your walkthrough comment](https://github.com/zenml-io/kitaru/pull/591#issuecomment-5015914816) in `cbd9ca2d`.

- The support agent now receives the complete `SupportDecision` JSON schema, has two bounded output retries, and gets the Pydantic validation details on each retry. The deterministic test exercises two invalid responses followed by a valid one.
- `escalate_to_human` now uses stable structured arguments: `customer_id` and a constrained `policy_label`. Human-readable audit text is derived inside the local tool implementation, so exact recorded-response matching no longer depends on reproducing model-written prose.
- The walkthrough fixture and source identity now use `v2.3-structured-escalation-imported`, and the README describes the verdict-led output readers should verify.

The acceptance journey now proves both sides of the contract: the exact structured call produces one recorded-response hit and `PASS`, while a different valid policy label is blocked with `argument_mismatch` and yields `HOLD`. The original write callable is never invoked.

Validation is green: `just check`, the focused replay suites, and the full suite (`4183 passed, 29 skipped`). I could not rerun the paid OpenAI path in this environment because `OPENAI_API_KEY` is not set, so the checked-in small fixture records its derived contract provenance explicitly.

htahir1:
Re-ran the full walkthrough at `e458e51a` from a clean checkout and a fresh isolated config — both findings are verified fixed, and every step now lands exactly as the README describes:

- **Step 6 reproduction** (previously failed on every attempt): `PASS account-setting-reproduction` — `recorded_path_comparable`, recorded replies **1/1 served, 0 missed** (the structured escalation matched and was served from the recording), 0 blocked, 0 divergences, objective 1.0, all protections passed.
- **Step 7 weakened counterfactual**: `HOLD` with `degraded` comparability, 1/4 served / 3 blocked / 1 divergence — the verdict engine correctly refuses comparability while still scoring the outcome.
- **Step 8 fix variant**: `PASS account-setting-fix`, clean exact-match continuation.
- **Steps 9–10**: two-case root-input suite reports `HOLD` as documented; the bounded rerun (`--max-trials 1 --max-cost-usd 1.00`) honors the caller idempotency key and lands `PASS`.

The simplified command surface (defaults + derived idempotency keys) and the version bump to `v2.3-structured-escalation-imported` for the regenerated fixture are both nice touches — the PASS → HOLD → PASS arc reads clearly now.

Two small unrelated notes while you're in the branch:

1. This branch's root `README.md` predates the #587/#589 rewrite on develop, so merging will revert it — `git checkout origin/develop -- README.md` (plus re-applying any intended additions) would avoid the regression.
2. `docs/book/adapters/pydantic-ai.md` and the adapter's in-repo README list `KitaruRunContext` among the exported lower-level wrappers, but it isn't exported from `kitaru.adapters.pydantic_ai` — either export it or drop it from both lists. (I've already removed it from the GitBook page on the `docs/vision-restructure` branch, which is now stacked on this one.)

PR #611, opened 2026-07-24
Add trace importers and PydanticAI adapter

AlexejPenner:
## Summary

This PR adds the first Kitaru v2 trace importer for Langfuse, basic evaluators, and a PydanticAI adapter.

The executable importer and evaluator implementations live under the repository-level `plugins/` tree rather than `src/kitaru`. This keeps provider code out of the core wheel and establishes the intended boundary for publishing official plugins as separate PyPI packages later. Standalone package metadata and a public facade are deferred.

The initial importer scope is limited to one provider. Braintrust and OpenTelemetry implementations live on the stacked [`codex/v2-importer-braintrust-otlp`](https://github.com/zenml-io/kitaru/compare/v2-importer...codex/v2-importer-braintrust-otlp) branch.

Until plugin packages are published, an explicit seeding script registers the repository implementations as inline PEP 723 plugins. Registration uses the configured public API client, does not run during server startup, and does not require separate adapter credentials.

### Langfuse importer

- Add the Langfuse JSON and JSONL importer at `plugins/importers/langfuse.py`.
- Accept trace rows, observation rows, and legacy ingestion-event exports.
- Emit `ParsedSession`, `ParsedNode`, and `ImportFailure` directly through the worker importer contract.
- Reconstruct parent-child graphs and isolate malformed sessions as import failures.
- Preserve external trace identifiers, model metadata, token usage, costs, timestamps, errors, and replay-readiness information.
- Enforce the 50 MiB upload limit before decoding.

### Importer boundary

- Keep the provider implementation and its parsing helpers outside `src/kitaru`.
- Use the worker's `ParsedSession`, `ParsedNode`, and `ImportFailure` models as the only result contract.
- Remove the additional provider-neutral model hierarchy from `src/kitaru/importers`.
- Validate node graphs inside the plugin, preserve valid sessions when another session fails, and attach bounded replay-readiness metadata.

### Evaluators

- Add cost, latency, and repeated tool-call evaluators at `plugins/evaluators/basic.py`.
- Add PEP 723 metadata for inline execution.
- Remove the evaluator implementations from the core package namespace.

### PydanticAI adapter

- Add `KitaruAgent` using PydanticAI capability hooks.
- Use the configured `KitaruAPIClient`, including worker task-token authentication and local persisted configuration.
- Record sessions, model requests and responses, tool calls, token usage, outputs, and failures.
- Apply task inputs, replay prompt and model overrides, and imported conversation history.
- Support passthrough, static, and history policies for local tools.
- Preserve synchronous, asynchronous, streamed, overridden, and concurrent runs.
- Keep the original agent exception primary if terminal Kitaru recording also fails.
- Add a runnable PydanticAI example.

### Development plugin registration

Add `scripts/seed_default_plugins.py` to register these repository implementations through the public Kitaru API:

- Importer: `langfuse`.
- Evaluators: `cost`, `latency`, and `tool-call-patterns`.

The script is idempotent and uses the generic server-side name filter:

- It creates a marked plugin parent when none exists.
- It uploads source and creates an immutable version only when the source digest or entry point changes.
- It refuses to modify an existing same-name parent that was not created by the seed flow.
- It does not replace a package-backed latest version, allowing published packages to take ownership later.

## Registering the plugins

Configure the default API client with `kitaru login`, or provide `KITARU_API_URL` and `KITARU_API_KEY`, then run:

```bash
export KITARU_PLUGIN_DISPLAY_VERSION="$(git rev-parse HEAD)"
uv run python scripts/seed_default_plugins.py
```

`KITARU_PLUGIN_DISPLAY_VERSION` is optional. Without it, the script uses an `inline-<source-digest>` label. Run the command once per workspace because plugin records and source blobs are workspace-scoped.

## Follow-up scope

Braintrust and OTLP are excluded from this PR. The stacked branch contains those two implementations, their tests, and seed registrations using the same `plugins/importers` layout.

Separate build metadata, dependency declarations, distribution names, and public facades for the plugin packages remain follow-up work. Langfuse does not need to be independently buildable in this PR.

## Design limits

- Import payloads are decoded and normalized in memory rather than streamed.
- Inline plugins import the worker result models and API enums from the installed Kitaru runtime, so the execution environment needs a compatible Kitaru version.
- The repository-level `plugins/` tree is covered by linting, type checking, and tests but is excluded from the Kitaru wheel.
- The PydanticAI adapter cannot intercept provider-native tools before execution and records them as provider-native observations.
- `LLMConfig` tool policies are not supported by this adapter and return an explicit error.
- Successful agent runs surface final Kitaru recording failures to the caller.

## Reviewer Notes

Start with `plugins/importers/langfuse.py` and the narrow parser result contract in `src/kitaru/task/importer.py`. The plugin now owns decoding, grouping, graph construction, failure isolation, and direct `ParsedSession` creation. A malformed trace should produce an isolated `ImportFailure` without discarding valid sessions in the same payload. Replay-relevant inputs, graph relationships, and provider metadata must survive direct construction of the worker models.

Review `scripts/seed_default_plugins.py` as the temporary deployment bridge. Its source paths should resolve under `plugins/`, and its ownership marker, exact-name lookup, digest comparison, and refusal to replace user-created or package-backed plugins should remain intact.

For the PydanticAI adapter, review prompt projection and replay overrides closely. Imported multi-turn conversations must retain prior history while replacing only the active prompt. Recording failures during an already-failing agent run must preserve the agent exception as the primary failure.

### Reproduction

1. Install the required extras.

   ```bash
   uv sync --extra server --extra pydantic-ai
   ```

2. Run the retained feature tests.

   ```bash
   uv run pytest -q \
     tests/adapters/pydantic_ai/test_adapter.py \
     tests/server/test_langfuse_importer.py \
     tests/evaluators/test_basic.py \
     tests/test_seed_default_plugins.py
   ```

3. Start a Kitaru server, configure the client, and run the seed script twice.

   ```bash
   uv run kitaru login --local
   uv run python scripts/seed_default_plugins.py
   uv run python scripts/seed_default_plugins.py
   ```

   The first run should report `created_version`. The second should report `unchanged`.

4. Configure `OPENAI_API_KEY` and agent identity as documented in the example, then run:

   ```bash
   uv run python -m examples.integrations.pydantic_ai_v2
   ```

Validation completed:

- `just check`
- Full suite: 2,376 passed, 450 skipped
- Focused importer contract tests: 32 passed


src/kitaru/worker/process.py
```diff
@@ -260,6 +261,35 @@ def parse_inline_dependencies(path: Path) -> list[str]:
     return list(dependencies)
 
 
+def filter_uninstalled_requirements(requirements: list[str]) -> list[str]:
```
schustmi: This fucks up some cases, and the whole packages are read from uv cache anyway after the first execution. Can you please tell your agents to not make any changes in the core code when implementing some importers or adapters. So far, nothing useful came from it.

src/kitaru/server/adapters/auth/auth_service.py
```diff
@@ -349,7 +349,10 @@ async def _resolve_control_plane_credential(self, credential: str) -> AuthContex
                 "Local API keys are rejected under control plane authentication."
             )
         context = await self._resolve_session_token(credential)
-        if context.account.external_id is None:
+        if (
```
schustmi: Remove, this is just wrong.

src/kitaru/server/application/services/default_plugins.py
```diff
@@ -0,0 +1,145 @@
...
+                name=name,
+                description=description,
+                provider=provider,
+                metadata={"built_in": True},
+            )
+        )
+        await repository.create_version(
+            importer.id,
+            PackagePluginSource(
+                requirement=f"kitaru=={KITARU_VERSION}",
```
schustmi: This by default means the users will always stay with some super outdated version of our importers, from the first version they used?

src/kitaru/server/api/app.py
```diff
@@ -195,9 +199,10 @@ async def lifespan(app: FastAPI) -> AsyncGenerator[None, None]:
                     repository=SQLAccountRepository(session),
                     password_hasher=BcryptPasswordHasher(),
                 )
-                await account_service.ensure_account(
+                account = await account_service.ensure_account(
                     settings.DEFAULT_ACCOUNT_NAME, settings.DEFAULT_ACCOUNT_PASSWORD
                 )
+                await ensure_default_plugins(SQLPluginRepository(session), account.id)
```
schustmi: So we only add plugins for non-control-plane servers? :D

schustmi (changes requested):

AlexejPenner:
@schustmi i have removed the offending workarounds - for now we simply have to seed the importers and evaluators into the db as inline plugins

schustmi (approved):

PR #641, opened 2026-07-31
Deploy kitaru to cloud.zenml.io

AlexejPenner:


src/kitaru/server/api/main.py
```diff
@@ -16,16 +16,31 @@
...
 from kitaru.server.api.app import create_app
 from kitaru.server.api.config import APISettings
 
 
+def app() -> FastAPI:
+    """Create the API application from environment settings.
+
+    Returns:
+        Configured FastAPI application.
+    """
```
schustmi, line 30: @stefannica The OTEL changes introduced the logging.basicConfig here in the main function, you think we should also run it here? Haven't tested whether its necessary, but prob still good to set at least the log level correctly?

stefannica: done

stefannica (approved):

PR #652, opened 2026-08-03
Add the Kitaru v2 CLI and native MCP server

### LINKED ISSUE #620: Kitaru v2 CLI (from scratch): login, worker, registration
Kitaru v2 ships a brand-new CLI, written from scratch (there is no current CLI to audit). Core surface from the MVP story:

- `kitaru login` (OAuth against your workspace; `--local` for self-hosted)
- worker lifecycle: start a local worker that polls the server and runs jobs (imports, replays, scoring)
- registration: agents, importers, scorers — the manual door for everything the skills do
- `kitaru worker add <provider>` (post-MVP remote workers; data model lands now)

Principle: everything a skill does must be doable by hand with the CLI. Two doors, same rooms.

**Priority:** P0 · **Target:** 2026-08-07

### LINKED ISSUE #624: New Kitaru MCP server
New MCP server for Kitaru v2, exposing the server's data model (agents, sessions, cohorts, experiments, scorers) and job triggers to coding agents — complementing the skills as a programmatic door.

**Priority:** P1 · **Target:** 2026-08-14

strickvl:
## Summary

Kitaru can now be operated through a full v2 CLI for humans, scripts, and coding agents, plus an optional native MCP server for MCP clients. This version deliberately keeps all new behavior inside the CLI and MCP packages and uses the existing `v2-spec-consolidated` server, SDK, API-model, worker, adapter, and OpenAPI contracts unchanged.

- The CLI covers explicit login, diagnostics, schema discovery, asset registration, session import and inspection, evaluations, cohorts, experiments, runs, workers, and jobs. Interactive terminals get curated views while JSON, JSONL, and machine text retain complete stable records.
- The MCP server exposes a capability-filtered surface of 2 read-only tools, 5 standard tools, and 7 destructive tools. MCP remains an optional package extra and calls the existing typed SDK directly.
- CLI connection selection follows `--server`, then `KITARU_API_URL`, then the server URL stored by `kitaru login`. MCP keeps an explicit process target selected from `--server`, `KITARU_MCP_SERVER`, or `KITARU_API_URL` at startup. `KITARU_API_KEY` overrides a stored credential for the selected URL.
- The CLI and MCP packages contain their own connection, reference-resolution, and redaction helpers. They do not add shared client utilities or require server feature discovery.
- MCP workflow creation is limited to blob-backed session import, which the existing client already supports. Replay, evaluation, session-run, and experiment-run creation remain CLI-only until the underlying API provides a contract Michael is comfortable supporting.
- Worker commands use the existing worker API and require `KITARU_API_KEY`; the CLI does not create a worker credential file or change worker lifecycle behavior.

There is no PR diff against the `v2-spec-consolidated` base under `openapi/`, `src/kitaru/adapters/`, `src/kitaru/api_models/`, `src/kitaru/client/`, `src/kitaru/server/`, `src/kitaru/task/`, or `src/kitaru/worker/`.

Related: #620

Closes #624

## Design decisions

- Human presentation is command-specific, but structured serialization always uses the complete command result so display choices cannot remove machine data.
- Finite results use JSON and long-running commands use ordered JSONL lifecycle events, with diagnostics kept on stderr.
- MCP tool availability is filtered before discovery. Read-only is the default, and local-path uploads, registration tools, resources, prompts, remote MCP transport, and persisted target contexts remain outside this release.
- List operations return one bounded page. Mutations require exact IDs or exact positive version numbers where supported by the existing SDK.
- Non-idempotent workflow starts that would need new server support are not emulated in the CLI or MCP layer.

## Validation

- Full test suite: 2,770 passed, 1 skipped.
- Focused CLI and MCP suites: 309 passed, 1 skipped.
- MCP schema budget passed with 2, 5, and 7 tools in read-only, standard, and destructive modes. Discovery sizes are 58,372, 83,962, and 93,344 bytes respectively.
- Ruff format and lint, repository-wide type checking, typos, YAML, actionlint, offline links, and `git diff --check` passed.

## Reviewer Notes

The central review question is now narrow: do the CLI and MCP packages provide useful frontends over the existing client without changing the contracts Michael is developing? The previous server-side idempotency implementation, shared connection/reference helpers, worker lifecycle changes, API feature advertisement, replay-config extensions, and associated migrations/tests have all been removed from this PR.

Please focus on complete JSON/JSONL output, explicit target and credential precedence, CLI partial-failure receipts, worker environment restoration, MCP capability filtering, strict request schemas, schema size, lifecycle cancellation/close ordering, and the boundary check below.

### Reproduction

1. Install development dependencies with `uv sync --extra cli --extra worker --extra mcp`.
2. Set `KITARU_API_URL` and, when required, `KITARU_API_KEY`, then exercise `uv run kitaru status`, `uv run kitaru agent list`, `uv run kitaru session list`, and one supported mutation.
3. Repeat representative CLI reads with `--output json`, `--output jsonl` where supported, and `--machine -o text`; confirm stdout contains complete machine-readable records and diagnostics remain on stderr.
4. Run `just mcp-schema-check`. Start `kitaru-mcp` in read-only, standard, and destructive modes and confirm discovery contains 2, 5, and 7 tools respectively.
5. Run `just build`, `just cli-artifact-smoke`, and `uv run --no-sync python scripts/smoke_mcp_wheel.py dist/kitaru-0.21.0-py3-none-any.whl`.
6. Confirm the core boundary with `git diff --name-only origin/v2-spec-consolidated...HEAD -- openapi src/kitaru/adapters src/kitaru/api_models src/kitaru/client src/kitaru/server src/kitaru/task src/kitaru/worker tests/client tests/server tests/task tests/worker`; it should print nothing.

This PR has no database migration, server rollout, or external data mutation. During release validation, install the built wheel in a clean environment, confirm the CLI reports the expected explicit target, confirm MCP startup writes redacted target/credential provenance only to stderr, and verify tool discovery for each capability mode. Treat unexpected authentication errors, protocol output on stdout, or schema-count drift as rollback signals. Roll back the PR merge or the release artifact; no data rollback is required. The PR author owns this validation during the first release candidate using the branch artifacts.


strickvl:
@schustmi dug into the plan (some details of which I had missed) and here are the reasons for the worker changes FWIW: https://gist.github.com/strickvl/f2c9c304be9f0c291d3cd4e3635ae5c0

Some of these def seem less defensible but I am not so aware of some of the auth decisions that you've been working on underneath, so maybe you can just give me a list of what can stay + what has to be reverted and I'll do that?

src/kitaru/server/domain/task.py
```diff
@@ -745,6 +745,7 @@ class ImportTaskDetails(FrozenModel):
     provider: str | None = None
     agent_id: uuid.UUID
     params: dict[str, Any] = Field(default_factory=dict)
+    tags: list[str] = Field(default_factory=list)
```
schustmi: this seems like a strange arbitrary addition which isn't connected to anything else? 🤔

src/kitaru/worker/heartbeat.py
```diff
@@ -75,16 +75,14 @@ async def run(self) -> None:
         while True:
             await asyncio.sleep(self._interval)
             task_ids = list(self._registered)
-            if not task_ids:
```
schustmi, line 78: This is wrong. If there are no tasks, the worker will try to claim new tasks, which the server treats as a heartbeat. In general, I'd suggest we remove any worker changes from this PR. Most of it is garbage.

schustmi (changes requested):

src/kitaru/worker/heartbeat.py
schustmi: If some CLI commands are broken because of some worker stuff not working on windows or we need additional worker changes, we should intentionally make those changes with a thought out plan in a separate PR, instead of a byproduct on some unrelated PR

AlexejPenner:
## Login succeeds but leaves subsequent commands without a server target

I hit this while running the canonical example against a hosted staging server.

### Reproduction

```console
$ kitaru login https://a7b333aa-kitaru.staging.cloudinfra.zenml.io
authentication   authenticated
credential_kind  device
context          -
context_active   false

$ kitaru worker start --name returns-example-worker
Error: Invalid configuration
No Kitaru server is configured.
Try: Pass --server, set KITARU_API_URL, or select a context.
```

The device credential is stored correctly. The problem is that `login SERVER` does not persist or activate the server as a target unless the user also supplies `--context`. Target resolution for the next process only considers an explicit `--server`, `KITARU_API_URL`, or an active context. The stored credential is keyed by server URL, but credentials alone cannot select a server.

This makes the expected onboarding flow fail:

```console
kitaru login SERVER
kitaru worker start
```

The login receipt reports success while leaving the CLI unusable until the user understands the separate context model.

### Proposed behavior

After a successful `kitaru login SERVER`, make `SERVER` the selected target for subsequent commands.

The UX contract should be:

- `kitaru login SERVER` authenticates and selects that server.
- `kitaru --context NAME login SERVER` authenticates, creates or updates `NAME`, and activates it, as it does today.
- Explicit `--server`, `--context`, and `KITARU_API_URL` keep their existing override precedence.
- Failed authentication must not change the selected target.
- If persisting an unnamed selected server does not fit the config model, create and activate a reserved default context internally. Users should not need a second `context add` and `context use` sequence after login.

The successful receipt should then show an active target and include a next action such as `kitaru status`.

### Tests

Please cover:

1. Login without `--context` selects the explicit server.
2. A fresh process can run `status` and `worker start` using the stored selection and credential.
3. Login with a named context still creates and activates that context.
4. Failed login preserves the previous selection.
5. Login to a second server updates the selected target without deleting credentials for the first server.

schustmi:
> ## Login succeeds but leaves subsequent commands without a server target
> I hit this while running the canonical example against a hosted staging server.
> 
> ### Reproduction
> ```
> $ kitaru login https://a7b333aa-kitaru.staging.cloudinfra.zenml.io
> authentication   authenticated
> credential_kind  device
> context          -
> context_active   false
> 
> $ kitaru worker start --name returns-example-worker
> Error: Invalid configuration
> No Kitaru server is configured.
> Try: Pass --server, set KITARU_API_URL, or select a context.
> ```
> 
> The device credential is stored correctly. The problem is that `login SERVER` does not persist or activate the server as a target unless the user also supplies `--context`. Target resolution for the next process only considers an explicit `--server`, `KITARU_API_URL`, or an active context. The stored credential is keyed by server URL, but credentials alone cannot select a server.
> 
> This makes the expected onboarding flow fail:
> 
> ```
> kitaru login SERVER
> kitaru worker start
> ```
> 
> The login receipt reports success while leaving the CLI unusable until the user understands the separate context model.
> 
> ### Proposed behavior
> After a successful `kitaru login SERVER`, make `SERVER` the selected target for subsequent commands.
> 
> The UX contract should be:
> 
> * `kitaru login SERVER` authenticates and selects that server.
> * `kitaru --context NAME login SERVER` authenticates, creates or updates `NAME`, and activates it, as it does today.
> * Explicit `--server`, `--context`, and `KITARU_API_URL` keep their existing override precedence.
> * Failed authentication must not change the selected target.
> * If persisting an unnamed selected server does not fit the config model, create and activate a reserved default context internally. Users should not need a second `context add` and `context use` sequence after login.
> 
> The successful receipt should then show an active target and include a next action such as `kitaru status`.
> 
> ### Tests
> Please cover:
> 
> 1. Login without `--context` selects the explicit server.
> 2. A fresh process can run `status` and `worker start` using the stored selection and credential.
> 3. Login with a named context still creates and activates that context.
> 4. Failed login preserves the previous selection.
> 5. Login to a second server updates the selected target without deleting credentials for the first server.

yes, we still need to figure out where to store credentials after login.

src/kitaru/server/domain/task.py
strickvl: > this seems like a strange arbitrary addition which isn't connected to anything else?

Fixed in f0292be9. I removed that task-domain addition and restored `src/kitaru/server/domain/task.py` to the `v2-importer` base. The PR now has no server-domain diff against its merge base.

src/kitaru/worker/heartbeat.py
strickvl: > I'd suggest we remove any worker changes from this PR.

Fixed in f0292be9. I removed the heartbeat, process, worker, configuration, and Python-entrypoint changes and restored `src/kitaru/worker/` to the `v2-importer` base. The CLI worker commands now adapt to the existing worker API and environment contract; any future worker behavior change can be designed and reviewed separately.

PR #659, opened 2026-08-04
Build canonical Example

AlexejPenner:
## Summary

- add the canonical returns-resolution walkthrough from trace generation through import, evaluation, investigation, cohorts, replay, and comparison
- merge the current `v2-spec-consolidated` importer, evaluator, and standalone adapter packages
- update the coding-agent walkthrough to use the current Kitaru MCP and CLI surfaces
- keep `kitaru-pydantic-ai` independently packaged and install it explicitly for the example
- document and test local candidate wheels for source checkouts before plugin versions are published
- include the OpenAI Agents and LangGraph standalone examples in the public example index and coverage manifest

## Validation

- `just check`
- `uv run pytest -q`: 2675 passed, 526 skipped
- canonical contracts: 20 passed
- canonical Docker E2E: 1 passed
- plugin source suite: 394 passed
- adapters repeated 10 times: 2010 passed
- installed plugin artifact smoke: passed
- trace corpus: 64 files, 78 sessions, 903 nodes, zero failures; 78-session canonical JSONL round trip passed

## Reviewer Notes

1. From the repository root, run `uv sync --frozen --extra cli --extra worker --extra examples --extra server --extra otel`.
2. Follow `examples/pydantic_ai_ticket_resolver/README.md` through the local wheel setup and server startup.
3. Run `uv run pytest -q tests/test_canonical_example.py` for the provider-free contract suite.
4. For the complete local workflow, run the E2E command listed in `examples/example-coverage.yaml` against an isolated local server.
5. Confirm `docker compose build server` succeeds without copying `plugins/` into the dependency layer.


strickvl:
I fixed the issues found while running through the canonical example:

- MCP list calls now return populated activity, registry, and investigation pages instead of failing with a generic internal error.
- The README now preserves `$280` correctly in the shell command.
- Replays use the same `gpt-5-nano` model as the checked-in baseline by default, so the latency comparison is meaningful unless `BASELINE_MODEL` is overridden.
- Missing replay cost is now reported as unavailable instead of appearing to be a real zero-dollar run.
- The checked-in traces and trace generator now give each imported session a clear per-ticket name.

I added regression coverage for all three MCP list families and for the example, trace names, model choice, README command, and missing-cost behavior.

strickvl:
@AlexejPenner I'd really like to merge this one in, though I see at least one change to core server code so we should check if that's really necessary.

strickvl:
The only `src/kitaru/server` change in this PR (the `EvaluatorConfig.evaluator` type in `replay_config.py`) is a server bug the canonical example exposed: the server registers default evaluators as `kitaru/cost` etc., but the replay config field rejected slashed names. Extracted it as #701 against `v2-spec-consolidated`; once that merges and this branch syncs, the server diff here disappears. (Also fixed in d21ca02d: the `v2-spec-consolidated` merge renamed `PluginName` → `NamespacedName`, which broke this branch's import.)

schustmi (approved):

PR #670, opened 2026-08-05
Introduce new fields, more importers and evaluators, release flow for external plugins

AlexejPenner:
## Summary

This PR establishes the v2 contract for imported sessions and nodes, adds production importers for the trace formats we have collected, introduces a portable Kitaru JSONL format, packages the official plugins for independent release, and adds the built-in deterministic evaluator catalog.

The core Kitaru changes are limited to the shared import contract, the session and node fields required by API clients, persistence and REST mapping for those fields, and default plugin registration. Provider decoding, joining, normalization, and evaluator implementations live in separately releasable packages under `plugins/packages`.

## Session and node contract

`ImportedSession`, `ImportedNode`, and `ImportedItem` are the public output contract for importers. They are also the input contract for the strict `kitaru-jsonl` importer, which lets producers emit Kitaru-compatible session JSONL without writing a provider adapter.

Sessions expose:

- complete `inputs` and `outputs`
- `framework` when source evidence identifies an agent framework without conflict
- source identity, timing, status, error, and metadata

Nodes preserve complete provider `inputs`, `outputs`, and `attributes`. These fields and `reasoning` are payload columns. The lightweight node response includes RFC 6901 JSON Pointer selectors for the human-readable input, output, and system prompt:

- `input_text_selector`
- `output_text_selector`
- `system_prompt_selector`

Importers set a selector only when they can identify one relevant string without guessing. `input_text_selector` and `system_prompt_selector` resolve against `inputs`; `output_text_selector` resolves against `outputs`. Clients can render the selected value after loading the payload and retain access to the complete source JSON for inspection. The session contract does not duplicate a system prompt from one of its nodes. Visible reasoning is retained when the source exports it as readable text; redacted, encrypted, or unavailable reasoning remains null.

## Import pipeline

Each provider importer performs the following stages:

1. Decode provider records into traces and spans.
2. Correlate related traces into one Kitaru session.
3. Order grouped traces as turns and reconstruct the node graph.
4. Normalize the shared session and node fields while retaining source payloads.

The import task validates the typed contract, creates the session, flattens nested node graphs into stable indexes, and writes nodes in batches. A malformed source group produces an isolated `ImportFailure`, allowing other valid sessions in the same upload to complete.

Multi-trace sessions use a versioned `turns` object in session inputs. Session outputs use the final turn output, with the latest node output as a fallback when the provider wrapper omits its result.

## Supported importers

| Importer | Package | Accepted input | Default grouping |
|---|---|---|---|
| `kitaru/langfuse` | `kitaru-langfuse-importer` | Trace, observation, and ingestion-event JSON or JSONL | `sessionId`, then trace ID |
| `kitaru/langsmith` | `kitaru-langsmith-importer` | Run-query and bulk-export JSON or JSONL | Known thread metadata, then trace ID |
| `kitaru/braintrust` | `kitaru-braintrust-importer` | Project-log and UI JSON exports | Known session or conversation fields, then trace ID |
| `kitaru/opentelemetry` | `kitaru-opentelemetry-importer` | OTLP collector envelopes, flattened OTLP JSONL, Arize JSONL, and Logfire JSONL | Standard conversation attributes, then trace ID |
| `kitaru/kitaru-jsonl` | `kitaru-jsonl-importer` | One strict Kitaru session object per JSONL line | One session per line |

Langfuse and LangSmith support user-selected correlation through `join_on`. Langfuse also accepts `join_path` plus `join_key`. Selectors use dotted paths or RFC 6901 JSON Pointers. An explicit selector must resolve to one non-empty value for every source trace. Missing or conflicting values become import failures instead of silently splitting a conversation.

Provider packages own their normalization rules. They retain source identifiers and graph relationships, status and errors, timestamps, model and provider fields, model parameters, token usage, cost, tool arguments and results, text selectors, visible reasoning, framework identity, normalization warnings, and source-completeness metadata.

## Official plugin packages and releases

Official plugins are distributed through six Python packages:

- five independently versioned importer packages listed above
- one `kitaru-evaluator` package containing all built-in evaluators

Each distribution publishes a catalog through the `kitaru.default_plugins` entry-point group. Server startup registers the catalog definitions as ownerless plugins under reserved `kitaru/` names and records the exact distribution requirement on every plugin version.

`plugins/default-requirements.txt` pins the package versions bundled into release server images. Importers can be released independently. All built-in evaluators share the `kitaru-evaluator` release cycle. The `Release plugin` workflow builds, tests, publishes, tags, and creates a GitHub release for one selected distribution. Local development uses the same catalogs through uv workspace members, without requiring a PyPI release.

The Kitaru release workflow verifies that every pinned plugin package exists on PyPI. The release server image installs those exact wheels and records them in its environment snapshot.

## Built-in evaluators

`kitaru-evaluator` contains the three existing basic evaluators and ten offline deterministic evaluators:

- basic signals: cost, latency, and tool-call patterns
- descriptive diagnostics: session integrity, trajectory signals, tool health, timing profile, and LLM-call signals
- configured policies: output contract, resource budget, tool policy, model policy, and workflow conformance

Descriptive evaluators report evidence without assigning a pass or fail verdict. Policy evaluators pass only when the session is terminal and the required evidence is complete and consistent. Insufficient evidence leaves `passed` unset. Results include input and configuration hashes, bound retained evidence, and hash oversized structured values before they reach the worker result limit.

Evaluators run only through explicitly started evaluation jobs. Importing or recording a session does not start them automatically.

## Corpus validation

The provider importers were exercised against the [Kitaru trace corpus](https://github.com/zenml-io/kitaru-trace-corpus), covering Langfuse, Langfuse SDK, Braintrust, LangSmith, Arize, and Logfire exports:

- 64 provider files
- 78 imported sessions
- 903 inspectable nodes
- 264 model calls
- 209 tool calls
- 21 nodes with visible reasoning
- 0 parser or selector-resolution failures

The audit resolves every emitted JSON Pointer against its retained payload and requires a non-empty string. It also checks session and node presence, framework detection, source payload retention, and isolated failures. Repository tests cover API round trips with and without payload loading.

## Reviewer Notes

The review has four contracts worth checking independently:

1. `src/kitaru/task/importer.py` and the session API models define the public producer and frontend contract.
2. Each importer package owns provider decoding, correlation, graph reconstruction, and normalization.
3. `kitaru-evaluator` owns the complete built-in evaluator catalog and its shared release cycle.
4. Server bootstrap, default requirement pins, Dockerfiles, and release workflows connect package versions to registered plugins and release images.

The main compatibility points are the JSON Pointer selector semantics, graph index validation, explicit join failure behavior, package-backed plugin versioning, and the fact that one `kitaru-evaluator` release advances all built-in evaluator definitions.

Plugin wheels must be published before a Kitaru release pins them. The release workflow and release Dockerfile both enforce this ordering.

### Reproduction

Install Kitaru with every workspace plugin and run repository validation:

```bash
uv sync --all-packages --all-extras
just check
just test
```

Build every plugin distribution:

```bash
for package in plugins/packages/*; do
  uv build --project "$package" --out-dir "/tmp/kitaru-plugin-dist/$(basename "$package")"
done
```

Inspect the discovered catalogs and owning requirements:

```bash
uv run python - <<'PY'
from kitaru.server.api.bootstrap import _load_default_plugin_definitions

for definition in _load_default_plugin_definitions():
    print(definition.name, definition.requirement, definition.entrypoint)
PY
```

For the node payload contract, import a node containing selectors and list it once with the default payload projection and once with `include_payloads=true`. Both responses include the selectors. The payload response also includes `inputs`, `outputs`, `attributes`, and visible reasoning.

Local validation on the merged tree:

- `just check` passes
- full suite: 2,730 passed, 515 skipped
- all six plugin wheels and source distributions build successfully


src/kitaru/server/application/services/session_node_service.py
```diff
@@ -155,6 +155,10 @@ async def ingest_nodes(
                 error=item.error,
                 started_at=item.started_at,
                 ended_at=item.ended_at,
+                input_text=item.input_text,
```
schustmi: All 4 of those (if they stay) need to be in `NodeRepository.PAYLOAD_COLUMNS` so they're not being loaded every time

schustmi (changes requested):

src/kitaru/server/domain/session.py
```diff
@@ -303,6 +304,14 @@ def update_name(self, name: str | None) -> None:
...
         self.name = name
 
+    def update_system_prompt(self, system_prompt: str | None) -> None:
+        """Set a new system prompt.
+
+        Args:
+            system_prompt: New system prompt.
+        """
+        self.system_prompt = system_prompt
+
```
AlexejPenner: not sure if this should be updatable

schustmi: i agree, it shouldn't be. sessions should be frozen apart from few things like outputs and status

strickvl:
I pushed commit `f987cf7f` with the fixes found during the real-trace review:

- Langfuse imports now recognize tool calls when newer exports store the function name in flattened `attributes.name` metadata. In the trace-corpus scenario, all 13 function spans now import as tool calls, including the failed call.
- Model policy checks now return HOLD when a session contains no LLM calls. Missing evidence is no longer treated as a policy pass.
- Cost-budget checks no longer require tool-call nodes to have cost metadata. They can now reconcile the recorded LLM costs with the session total and return a real pass or fail.

The earlier `modelId` problem had already been fixed by the preceding teammate commit, so I left that code unchanged.

I added regression tests for all three cases. The focused importer and evaluator suites pass with 78 tests. The full suite reached 3,245 passing tests and 1 skipped test; the remaining 2 failures are unrelated OpenTelemetry tests caused by missing instrumentation packages in this local environment.

docker/dev-client.Dockerfile
```diff
@@ -66,7 +67,7 @@ USER $USERNAME
 
 # Install the repository project and its client dependencies from the lockfile.
 # Keep a snapshot of the resulting environment for external inspection.
-RUN uv sync --locked --no-dev --no-editable && \
+RUN uv sync --locked --package kitaru --no-dev --no-editable && \
```
schustmi: There are new docker files on the base branch, rebase + also apply there.

docker/release-client.Dockerfile
```diff
@@ -60,15 +66,25 @@ ENV UV_COMPILE_BYTECODE=1 \
...
+# the resulting environment for inspection outside the container.
 RUN test -n "$KITARU_VERSION" && \
   test "$(uv version --short)" = "$KITARU_VERSION" && \
-  uv sync --locked --no-dev --no-install-project && \
-  uv pip install \
-    --no-deps \
-    --only-binary=:all: \
-    "kitaru==$KITARU_VERSION" && \
+  uv sync --locked --no-dev --no-install-workspace && \
+  if [ "$PACKAGE_SOURCE" = candidates ]; then \
```
schustmi: What is this? Pretty sure this has no reason being in our release images.

In general, the whole point of the plugins being separate packages is that nothing in these images should even be touched by your PR

scripts/seed_default_plugins.py
```diff
@@ -1,257 +0,0 @@
-#!/usr/bin/env python3
```
schustmi, line 1: Wouldn't this script still make sense. If I'm developing new plugin versions locally, how would I register them in the server? Or I guess it's just a single CLI command anyway, and that's why we don't need it?

schustmi (changes requested):

src/kitaru/server/api/bootstrap.py
```diff
@@ -13,34 +13,31 @@
...
     DuplicatePluginVersion,
+    PackagePluginSource,
     Plugin,
     PluginKind,
     PluginNotFound,
-    ScriptPluginSource,
 )
 
-_SOURCE_MEDIA_TYPE = "text/x-python"
+_DEFAULT_PLUGIN_ENTRY_POINT_GROUP = "kitaru.default_plugins"
```
schustmi: This makes 0 sense imo, hard overengineering and even requires those packages in the server image/environment for no reason. Simply hardcoded list of `DefaultPluginDefinition`

schustmi (changes requested):

scripts/seed_default_plugins.py
AlexejPenner: kitaru importer register dev-langfuse \
  --script path/to/importer.py \
  --entrypoint parse \
  --provider langfuse \
  
  I would throw everythign out now that isnt really necessary, happy to build out better developer experience later

docker/release-client.Dockerfile
AlexejPenner: all these changes were meant to let me test with local wheels instead of published wheels in pypi  - moving it to separate docker files

schustmi: Maybe gitignored ones, because no one will need this once we verified it works once

schustmi (approved):

PR #698, opened 2026-08-08
Replays without an adapter

schustmi:
This PR adds replays that run without an adapter, through a function run spec on the agent version. Create the version with a run spec pointing at a Python function, as `module:attribute`. The function runs the agent in your own system and returns the external id of the resulting session. Kitaru records a placeholder session, and a later import of that external id adopts it, runs the evaluations against the imported content, and settles the replay and experiment run as usual. An imported run that itself failed fails the replay with that run's error.

**Example:**
```python
client.agents.create_version(
    agent_id,
    AgentVersionCreateRequest(
        run_spec=FunctionRunSpec(entrypoint="my_service.hooks:run_replay")
    ),
)
```

**Limitations:**
- Function entrypoints load installed Python packages, not uploaded script plugins.
- A replay whose import never arrives stays open until it is canceled.
- Run specs are now a discriminated union, so agent version create requests must carry `"type": "command"` or `"type": "function"`.

`typing-extensions`: Needed for nested discriminated unions for pydantic. Was already a hard requirement by pydantic, so we don't actually add additional dependencies to our client package, we just specify it explicitly now.


strickvl:
@schustmi Heads-up: I ran an automated multi-agent review over this branch, so treat these as *candidate* bugs that need human confirmation. That said, both survived an adversarial verification pass (one was reproduced against the PR's own test fixtures), and they share a root cause: the placeholder session, the agent task's completion, and the trace import are three independent writers, and only the ordering "placeholder → completion → import" is exercised by the tests. The other orderings look broken.

### 1. Import finalizes before the agent task's completion lands → replay spuriously FAILED

`src/kitaru/server/application/services/replay_pipeline.py` (`append_result_evaluations` + `complete_import_wait`)

- The driver creates the `PENDING_IMPORT` placeholder and exits; before the worker's COMPLETED update for the agent task commits, the import adopts and finalizes the session.
- `complete_import_wait` completes the `ImportWaitTask` (no check that the producing agent task is terminal), which runs `append_result_evaluations`: evaluator tasks are inserted and the replay moves PENDING → EVALUATING.
- The agent task's COMPLETED update then lands and `append_result_evaluations` fires a second time. The session is no longer `PENDING_IMPORT`, so the deferral guard passes; the second insert hits the unique constraint on `(job_id, input_session_id, plugin_version_id)` → `DuplicateEvaluationTask` (with zero evaluators it's `IllegalReplayStatusTransition` from `start_evaluating()` instead). Since event handlers run on the caller's transaction, this rolls back the completion.
- The worker resolves the 409 via `_resolve_completion_conflict` → marks the agent task FAILED → `on_failure=ABORT` cancels the just-created evaluator tasks → the replay lands FAILED even though run, import, and evaluations all succeeded, with an unrelated error message.

The verification agent reproduced the second-invocation conflict using the existing fixtures in `tests/server/test_replay_pipeline.py`. The window is normally sub-second, but a run function that returns with the trace already complete (synchronous external agent) widens it, and batch fan-out multiplies the trials. Possible fixes: make `append_result_evaluations` idempotent (skip when the replay already left PENDING), or have `complete_import_wait` defer until the producing agent task is terminal and let the agent-completion path finish the wait task when the session is already finalized.

### 2. Requeued function-agent attempt orphans the placeholder → retry blocked and/or import silently swallowed

`src/kitaru/server/application/services/task_service.py` (`_unlink_result_session`, ~line 479)

- Worker dies after the placeholder create but before completion is reported. The staleness sweep requeues the agent task; `_unlink_result_session` clears `session.task_id` but leaves the placeholder in `PENDING_IMPORT` with its `external_id` (the state `tests/server/test_task_service.py` around line 480 asserts).
- If the run function is idempotent per `replay_id` (which passing `replay_id` into the run function seems to encourage), attempt 2 returns the same external id and the placeholder create hits `uq_session_owner_id_external_id` → `DuplicatePendingImportSession` → `AgentRunError` on every retry → replay fails.
- Independently, when the real trace arrives it adopts the *orphaned* placeholder by `(owner, external_id)`; its `task_id` is `None`, so `complete_import_wait` returns at the `session.task_id is None` check, the wait task never completes, and it dies at the `import_deadline_seconds` sweep (default 24h). A successful external run silently becomes a failed replay a day later.

Possible fix: on requeue, expire/fail the orphaned placeholder, or make placeholder creation adopt an existing pending placeholder with the same external id and relink `task_id` so retries are idempotent.

---

One smaller, unverified note from the same sweep: an import that is processed before the placeholder insert lands (importer beats the driver's `sessions.create`) is not adopted at all, which ends in the same orphan/deadline outcome as bug 2's second half.

PR #699, opened 2026-08-08
Fix v2 importer normalization and grouping

AlexejPenner:
## Summary

- update default importers, the PydanticAI adapter, and deterministic evaluators for the consolidated `model_provider` node field
- keep `output_text_selector` on visible assistant text by excluding reasoning, thinking, and tool-call parts
- treat redacted Langfuse session identifiers as missing so unrelated traces do not merge
- add `kitaru session import --join-on /json/pointer` and pass the pointer through import-task parameters
- support scalar JSON Pointer grouping in Braintrust, Langfuse, LangSmith, and OpenTelemetry imports, including grouping provenance in session metadata
- document the CLI, SDK, REST, RFC 6901 escaping, provider pointer roots, compatibility behavior, and failure semantics
- bump the changed default plugin packages to 0.1.1 and update catalog pins

## Root cause

The consolidated v2 node model renamed `provider` to `model_provider`, but the plugin workspace still constructed and read the old field. The shared importer normalization pattern also traversed any textual content in assistant parts, so tool-only model turns could point `output_text_selector` at reasoning content. Langfuse redaction placeholders were accepted as real session identifiers, grouping separate traces under the same placeholder.

Explicit session grouping was available only through raw importer parameters, and support differed by provider. Langfuse and LangSmith accepted `join_on`; Braintrust and OpenTelemetry could use only their built-in session or conversation fields. The CLI had no discoverable option for this workflow.

## Validation

- `just check`
- `uv run --extra cli --extra server pytest -q tests/cli/test_sessions.py tests/server/test_imports_api.py tests/task/test_importer.py` (43 passed)
- `uv run --project plugins ruff format --config plugins/pyproject.toml --check plugins`
- `uv run --project plugins ruff check --config plugins/pyproject.toml plugins`
- `uv run --project plugins ty check --project plugins`
- `uv run --project plugins pytest -q -c plugins/pyproject.toml plugins/tests tests/server/test_default_plugins.py` (194 passed)
- `just cli-artifact-smoke`
- `just plugin-artifact-smoke`
- synthetic trace corpus audit: 104 artifacts, 128 sessions, 1,302 nodes, zero parse failures, zero selector resolution failures, zero output selectors targeting reasoning/tool parts, exact Kitaru JSONL round trip
- Markdown links and typos pass through `just check`

`just docs-validate` is unavailable on this branch because `docs/package.json` has no `validate:export` script. `just docs-build` could not run in the isolated worktree because `docs/node_modules` is not installed. The changed page is GitBook Markdown and passed the repository link and typo checks.

## Reviewer Notes

The output selector policy recognizes provider discriminator fields named `type`, `part_kind`, `kind`, and `event.name`. It excludes normalized values for reasoning, thinking, tool calls, tool use, and function calls. Review whether this exclusion list matches the visible-output contract intended for all four provider importers.

The `--join-on` CLI option accepts RFC 6901 JSON Pointer syntax and writes it to `ImportCreateRequest.params["join_on"]`. The server persists that dictionary on the import task, the worker includes it in `ImportTaskDetails.params`, and the task process calls `parse(payload, params)`. Direct SDK and REST callers use the same parameter. Existing dotted `params.join_on` values and Langfuse `join_path` plus `join_key` remain supported for compatibility.

Pointer roots differ by provider and are documented in `docs/book/guides/importing-sessions.md`. Selected values must be non-empty scalars. Missing, conflicting, object, or array values isolate the affected trace as an import failure.

The package bumps cover every changed default plugin distribution. The Kitaru JSONL importer remains at 0.1.0 because its implementation did not change.

### Reproduction

1. Check out this branch and sync with `uv sync --frozen --extra cli --extra worker --extra server --extra mcp --extra otel` and `uv sync --project plugins --frozen --all-packages`.
2. Run `kitaru session import PAYLOAD --importer kitaru/langfuse@latest --agent AGENT@latest --join-on '/metadata/customer/case_id' --wait` against an export containing multiple trace IDs with the same selected value.
3. Confirm the result contains one session with ordered turns and `langfuse.join_paths` records the pointer.
4. Run the validation commands above before merging.


strickvl:
@AlexejPenner ran a deep review of this branch (importer code read in full, live probes against the real `parse()` entrypoints, plugin + CLI/server suites re-run locally). Overall: no show-stoppers, the rename is complete, the `--join-on` plumbing is sound end to end, and the failure-isolation semantics behave as documented. Two small gaps sit in exactly the feature this PR ships (the visible-output filter), so it would make sense to fix them here before merge. The rest are fine as follow-ups.

### Worth fixing in this PR

**1. Gemini reasoning parts bypass the visible-only filter.** Gemini marks reasoning with a boolean flag rather than a string kind: `{"text": "...", "thought": true}`. The exclusion check in `_content_match` only inspects string values under `type` / `part_kind` / `kind` / `event.name`, so the part passes as visible. Reproduced live: for a thought-only model turn (thought part + `functionCall` part, no visible answer), `_output_text_selector` returns `/parts/0/text`, storing the hidden chain of thought as the node's visible output. Mixed turns are unaffected because `matches[-1]` picks the trailing answer part. Suggested fix: also treat `value.get("thought") is True` as non-visible, in all four copies of the helper (braintrust, langfuse, langsmith, opentelemetry importers).

**2. Anthropic `redacted_thinking` blocks slip through.** `{"type": "redacted_thinking", "data": "<base64>"}` normalizes to `redacted-thinking`, which is not in the excluded set, and `data` is in the recursion key list, so a `[redacted_thinking, tool_use]` turn yields a selector pointing at the encrypted blob (`/content/0/data`) instead of `None`. Reproduced live. One-line fix: add `redacted-thinking` to the excluded set in the same four places.

### Fine as follow-ups (flagging so they are a decision, not an accident)

- **Malformed `join_on` failure scope differs by provider.** The OTel importer resolves `_path_value` outside the per-trace try (around `importer.py:1439`), so an invalid escape or whitespace-only pointer fails the whole import task, while braintrust/langfuse/langsmith isolate it as per-trace failures. Loud either way, just inconsistent for the same user mistake.
- **The redaction guard (`_is_missing_join_value`) is Langfuse-only.** An OTel/Braintrust/LangSmith export whose session or conversation id was scrubbed to a literal `[REDACTED]` still merges unrelated traces into one session. Pre-existing behavior, but it is the same failure mode this PR fixes for Langfuse.
- **`_path_value` array tokens are laxer than the RFC 6901 the docs promise.** `int(part)` accepts `-1`, `+1`, `" 1"`, `1_0`; `/turns/-1/case_id` silently resolves to the last element instead of failing, so a typo can group by an unintended value with no error. Rejecting non-digit tokens (or documenting the leniency) would close it.
- **`--join-on` with the `jsonl` importer is a silent no-op.** The CLI merges it into `params` for any importer, and the jsonl importer discards params, so the user believes regrouping happened. Arguably inapplicable for that format, but a warning or error for unsupported params would avoid the confusion.
- **The documented failure semantics are untested for Braintrust and OTel.** No test feeds a dict/list join value and asserts an `ImportFailure`, and the missing-value case is only covered for Langfuse and LangSmith. The branches behave correctly today (verified by live probes), but nothing pins them.

Also confirmed, so nobody re-litigates them: the 0.1.1 pins follow the pins-first-then-publish process and `release.yml` gates on publication; the Braintrust default-grouping change to root-record based is an improvement (the old per-record path could split one trace across sessions), though it is worth a CHANGELOG mention.

strickvl:
Verified every observation in the review against the current PR head and the real importer helpers.

### Fixed in `4d15f42b`

- Gemini `{"thought": true}` parts are now excluded from visible output selection in all four provider importers.
- Anthropic `redacted_thinking` blocks are now excluded, so encrypted/redacted data cannot become `output_text_selector`.
- JSON Pointer list traversal now enforces RFC 6901 array-index grammar. Negative, signed, whitespace-padded, leading-zero, and underscored indices no longer silently select a Python list element. Numeric-looking object keys remain valid.
- Added shared regression coverage across Braintrust, Langfuse, LangSmith, and OpenTelemetry and updated the changelog.

### Confirmed, left as follow-ups

- Malformed `join_on` configuration has inconsistent task-wide versus per-trace failure scope. This needs one explicit configuration-error contract rather than a local exception-placement change.
- Redaction-placeholder handling is Langfuse-only. Extending it needs a provider-aware placeholder policy so legitimate identifiers are not discarded.
- `--join-on` is ignored by the Kitaru JSONL importer. Fixing this cleanly affects importer capability validation and the independently versioned JSONL package.
- Braintrust/OpenTelemetry failure-branch coverage and the Braintrust root-record grouping changelog note are real gaps, but the probed runtime branches behave correctly and are separate from the defects fixed here.

### Validation

- `just check`
- `just test`: 3,177 passed, 1 skipped
- plugin format, lint, and type checks
- full plugin suite: 394 passed
- focused importer suite: 123 passed
- simplify pass: no reuse, quality, or efficiency issues found

strickvl (approved):

PR #704, opened 2026-08-10
Remove AnnotationSelector.part

schustmi:
Remove the `part` field and `AnnotationSelectorPart` enum from `AnnotationSelector`, and validate that `span` can only be set together with `path`.


AlexejPenner (approved):
:seal:

PR #706, opened 2026-08-10
Add TypeScript canonical returns example

strickvl:
## What changed?

TypeScript users can now run the canonical ten-ticket returns walkthrough with AI SDK 7 and Kitaru's Vercel adapter. The example records sessions directly into Kitaru, turns the reviewed policy into the existing Python evaluator ABI, and replays exact target and control cohorts through a stricter agent version.

The example is a standalone package under `v2_examples/vercel_ai_ticket_resolver`. It does not change the root workspace, lockfile, scripts, or existing examples. This keeps the PR isolated on top of `feat/ts-support`, but root aggregate pnpm commands will not discover the example automatically.

## Why?

The shipped canonical walkthrough only demonstrated the Python/PydanticAI path. This gives Vercel adapter users the same use case and evidence-first flow without requiring product changes outside the new example.

The provider-free default uses `MockLanguageModelV4`. Its model identity and token counts are scripted integration evidence, not a claim that a real model improved. A paid OpenAI path exists behind both an explicit approval flag and a provider key.

## Reviewer Notes

Start with `README.md` for the manual path or `README_AGENT_GUIDED.md` for the MCP-guided path. Both preserve the canonical ten fixtures, six snake-case tools, 8-pass/2-fail baseline, reviewed target/control split, and five passing strict replays.

The main design decision is direct adapter recording instead of importing the Python Langfuse corpus. The Vercel adapter requires string task inputs and records a generation summary, so the documented Python evaluator parses `session.outputs.text` while retaining compatibility with the imported Python output shape.

The baseline runner commits a ticket-to-session manifest after each ticket. It resumes missing tickets, archives prior evidence on `--fresh`, adopts only an exact inspected completed orphan, and archives an exact inspected failed orphan before `--retry` records a new session. It never deletes remote sessions.

## Reproduction

Use Node 22 and run from `v2_examples/vercel_ai_ticket_resolver`:

```bash
pnpm --dir ../.. --filter @zenml-io/kitaru build
pnpm --dir ../.. --filter @zenml-io/kitaru-vercel-ai build
pnpm --ignore-workspace install --frozen-lockfile
pnpm build
pnpm test
pnpm typecheck
pnpm lint
docker compose -f ../../docker-compose.yml up -d --build
pnpm test:e2e
```

The unit suite should report 49 passing tests. The E2E proof should report one passing workflow test. That workflow records ten baseline sessions, asserts eight evaluator passes and failures only for tickets 004 and 007, creates the exact 2-ticket target and 3-ticket control cohorts, and asserts all five strict replays and replay evaluations pass.

## Local checks run

- Node 22 build, 49 Vitest tests, typecheck, and Biome lint
- Provider-free adapter smoke command with structured refund output
- Documented evaluator contract: 8 pytest tests passed
- Live PostgreSQL API/evaluator/cohort/replay proof: 1 pytest test passed in 9.12s
- Existing Vercel adapter integration test: 1 pytest test passed
- `git diff --check`

No service or production runtime changes in this PR require deployment monitoring. This is a source-checkout-only local example. After merge, validate it by running the reproduction path against the built local TypeScript packages and confirming the deterministic 8/2 baseline and five passing replays.


AlexejPenner (approved):
seal

PR #715, opened 2026-08-11
Gate the CSRF check on the session token contents

schustmi:
This PR requires the CSRF token whenever the resolved session token carries one, instead of whenever the credential arrives in the auth cookie.

- Tokens issued to cookie sessions now require the `X-CSRF-Token` header regardless of how the credential is sent.
- Tokens without an embedded CSRF token, such as those from the API key or device grants, never require one.


AlexejPenner (approved):
🦭

PR #716, opened 2026-08-11
Move investigation questions to sessions, add highlights

schustmi:
Removes investigation-level questions and the annotation question key. Each investigation session now carries an optional `question` and a list of `highlights`, each a selector plus a description, both set at investigation creation.


AlexejPenner (dismissed):
🦭

PR #721, opened 2026-08-12
Track user enriched event on finished onboarding survey

schustmi:
Track the "User Enriched" analytics event when a user first finishes the onboarding survey, keyed on the same `finished_onboarding_survey` metadata key as ZenML.


AlexejPenner (approved):
🦭

PR #723, opened 2026-08-12
Automate package release planning

AlexejPenner:


AlexejPenner:
Superseded by the staged release-system plan starting with #724. The Notion procedures now follow the independent Python package and coordinated bundle model.

PR #727, opened 2026-08-13
Allow filtering sessions by experiment run

schustmi:
Filter session lists with `experiment_run_id`, which matches only the sessions produced as the results of the run's replays.


AlexejPenner (approved):
:seal:

PR #728, opened 2026-08-13
Do not apply replay overrides to agent task inputs

schustmi:
Agent tasks created for replays now carry the raw baseline session inputs, the prompt and system prompt overrides are no longer baked in at task creation.


AlexejPenner (approved):
🦭

PR #760, opened 2026-08-14
Separate core and plugin release workflows

AlexejPenner:
## Summary

- make `.github/workflows/release.yml` own the complete core Kitaru release: PyPI, Docker Hub, managed ECR, OCI Helm, stable `latest` aliases, and the GitHub Release
- reserve `.github/workflows/release-plugins.yml` for plugin package publication
- route `python/kitaru/v*` tags to the core workflow and exclude them from the plugin workflow
- remove the `workflow_run` handoff and create immutable GitHub Releases after required artifacts publish

## Impact

The core and plugin release paths now have separate entry points and explicit ownership. Manual dispatch remains build-only for both workflows. Managed ECR publication remains non-blocking and reports a warning on failure.

Moving core PyPI publication from `release-plugins.yml` to `release.yml` requires the Kitaru PyPI Trusted Publisher workflow mapping to be updated when this PR merges. The existing `pypi-kitaru` GitHub environment remains in use.

## Validation

- formatting and Ruff lint passed
- type checking passed
- YAML validation and Actionlint passed
- link and spelling checks passed
- `uv run pytest tests/scripts/test_release_units.py` passed, 40 tests
- `just zizmor` reported no findings

`just check` currently stops at the OpenAPI gate because current `develop` has `pyproject.toml` at `0.22.0rc4` while `openapi/openapi.json` contains `0.22.0rc3`. This PR does not modify either file.


schustmi (dismissed):

AlexejPenner:
for rc5 i would like to already run with this so we already test this before tuesday

schustmi (approved):

PR #765, opened 2026-08-14
Serve the bundled Kitaru UI

schustmi:
The server serves the bundled Kitaru UI at its root URL with SPA fallback and immutable asset caching, redirects to the configured dashboard URL instead when `KITARU_SERVER_EXTERNAL_UI=true`, and reports the served UI version on `GET /v1/info`.


strickvl (dismissed):

strickvl (approved):

AlexejPenner (approved):
:seal:

PR #770, opened 2026-08-14
Fix local RC image tag

AlexejPenner:
## Summary

- Add a generic PEP 440 to Docker image-tag formatter based on the canonical version suffix.
- Convert prerelease, post-release, development, local, and combined suffixes without branching on individual `Version` fields.
- Keep published-image selection policy separate from version formatting.

## Root cause

The local runtime used the installed Python package version unchanged as the Docker image tag. PEP 440 uses compact suffix syntax, while deployable image tags use a hyphen and dotted suffix components.

## Reviewer Notes

`_format_image_version` supports canonical PEP 440 suffixes generically, including combined forms. `_get_server_image` continues to require `KITARU_LOCAL_IMAGE` for development and local builds, so self-built images remain an explicit selection and do not trigger registry pulls by default.

### Reproduction

1. Install the `0.22.0rc5` CLI package.
2. Run `kitaru login --local`.
3. Confirm Docker pulls `zenmldocker/kitaru-server:0.22.0-rc.5` instead of `zenmldocker/kitaru-server:0.22.0rc5`.

Validation: `uv run pytest tests/cli/test_local_runtime.py` and `just check`.

### DIFF
--- a/src/kitaru/cli/local_runtime.py
+++ b/src/kitaru/cli/local_runtime.py
@@ -397,7 +397,10 @@ def _get_server_image(package_version: str) -> tuple[str, bool]:
             "No published local server image is available for this development build.",
             hint=f"Set {LOCAL_IMAGE_ENV} to a compatible local image.",
         )
-    return f"zenmldocker/kitaru-server:{package_version}", False
+    image_version = package_version
+    if parsed_version.pre is not None and parsed_version.pre[0] == "rc":
+        image_version = f"{parsed_version.base_version}-rc.{parsed_version.pre[1]}"
+    return f"zenmldocker/kitaru-server:{image_version}", False
 
 
 async def _ensure_image(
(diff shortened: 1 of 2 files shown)


src/kitaru/cli/local_runtime.py
```diff
@@ -397,7 +397,10 @@ def _get_server_image(package_version: str) -> tuple[str, bool]:
             "No published local server image is available for this development build.",
             hint=f"Set {LOCAL_IMAGE_ENV} to a compatible local image.",
         )
-    return f"zenmldocker/kitaru-server:{package_version}", False
+    image_version = package_version
+    if parsed_version.pre is not None and parsed_version.pre[0] == "rc":
+        image_version = f"{parsed_version.base_version}-rc.{parsed_version.pre[1]}"
+    return f"zenmldocker/kitaru-server:{image_version}", False
```
schustmi: This should be generic, not some `==rc`.

If it has anything after the patch version, it needs to make those changes.

schustmi (changes requested):

schustmi (dismissed):

PR #782, opened 2026-08-14
Add framework and call count session analytics fields

schustmi:
This PR adds the framework, adapter version, LLM call count and tool call count to the session completed analytics event.


AlexejPenner (approved):
🦭

PR #783, opened 2026-08-14
Refresh the returns investigation example

AlexejPenner:
## Summary

- Consolidate the PydanticAI returns resolver guidance into one README with a skill-guided path and a manual path.
- Show the current Kitaru investigation workflow, including evidence-led investigations, verdicts, and line-targeted annotations.
- Remove scenario labels and answer-key metadata from public prompts, fixtures, and imported traces.
- Move deterministic test oracles into test-only fixtures and update the canonical example coverage.

## Impact

Users get one current workflow for importing and investigating the example. Public artifacts no longer leak expected outcomes that could bias an investigation.

## Validation

- `just check`
- `uv run --extra server --with-editable plugins/packages/pydantic-ai --with pytest pytest -q tests/test_canonical_example.py`
- `just example-coverage-audit`
- `just links`
- `git diff --check`

The targeted canonical example suite passed with 15 tests. The opt-in server-backed E2E test was not run because it requires an isolated Kitaru server and `KITARU_CANONICAL_*` configuration.

## Reviewer Notes

Review the example as a first-time user. Check that the skill-guided and manual paths reach the same investigation outcome without exposing expected answers in public artifacts.

### Reproduction

1. Follow `examples/pydantic_ai_ticket_resolver/README.md` from setup through trace import.
2. Install the recommended skills with `npx skills add zenml-io/kitaru-skills` and use the provided investigation prompt.
3. Follow the manual path to create an investigation, record verdicts, and add line-targeted annotations.
4. Run the validation commands above.


examples/pydantic_ai_ticket_resolver/tests/canonical_returns_agent.py
```diff
@@ -0,0 +1,123 @@
+"""Provider-free candidate agent for the canonical end-to-end contract."""
```
schustmi, line 1: Does it make sense to move these to a separate test dir in the example maybe?

AlexejPenner: yes it does

examples/pydantic_ai_ticket_resolver/README.md
```diff
@@ -143,364 +139,425 @@ uv run kitaru session list \
...
+agreement and limitations against the reviewed evidence.
+
+If I want to test a change, continue with the kitaru-replay-experiment skill.
+Show one complete run card with the exact cohort version, candidate agent
+version, evaluator versions and parameters, adapter support, explicit tool
+policy, expected model work, cost uncertainty, and claim limits. Ask before
+remote writes, candidate code changes, or paid replay. Supervise the accepted
+run to a terminal state and report exact paired case evidence as improved,
+regressed, trade-off, or inconclusive. Leave the deployment decision to me.
+```
```
strickvl, line 189: This prompt feels sort of aggressively large...

Why do we even need the automatic agent-driven journey? Baris also was against this kind of idea.

examples/pydantic_ai_ticket_resolver/README.md
```diff
@@ -143,364 +139,425 @@ uv run kitaru session list \
...
+
+The two named skills divide the work cleanly:
+
+- `kitaru-investigation` discovers behavior, stores human evidence, confirms a cohort, and selects or creates an evaluator.
+- `kitaru-replay-experiment` tests one accepted candidate against exact cohort and evaluator versions.
+
+The skills will pause for human judgments and consequential writes. This is part of the evidence model. An assistant suggestion never becomes a human annotation or verdict without your confirmation.
+
+## Manual route: operate the evidence loop yourself
```
strickvl, line 198: Should mention somewhere in CLAUDE.md / AGENTS.md that this example is very closely tied to the quickstart (whatever form it ends up taking) and that updates to one or the other should require similar updates to the corresponding parallel one etc..

docs/book/getting-started/quickstart.md
strickvl, line 1: I've since rewritten the quickstart in my PR quite a bit and split it up etc... We're going to have to closely coordinate who merges their PR first.

strickvl (changes requested):

strickvl (approved):

PR #784, opened 2026-08-14
Stream local Docker startup progress

AlexejPenner:
## Summary

Stream Docker image-pull and Compose startup output during interactive `kitaru login --local` runs instead of displaying a static progress sentence while Docker works in the background.

Interactive startup now forwards Docker output to stderr as it arrives and requests plain Compose progress output. Non-interactive and machine-oriented runs continue capturing Docker output so their structured output remains clean. Streaming commands retain bounded timeouts and command-specific failure messages.

Add regression coverage for interactive pull and Compose streaming, the silent non-interactive path, and the existing captured execution path.

## Reviewer Notes

PR #780 improved the image-conflict error, but the progress line shown before local startup could remain unchanged for the entire image pull and container creation. This follow-up exposes the underlying Docker progress without changing the structured CLI contract.

### Reproduction

1. Remove the expected local server image or choose a new valid image through `KITARU_LOCAL_IMAGE`.
2. Run `kitaru login --local --upgrade` interactively.
3. Confirm Docker pull and Compose startup lines appear as they happen.
4. Run with `--non-interactive --output json` and confirm Docker progress does not corrupt structured output.

Validation: 39 focused auth and local-runtime tests passed. Focused Ruff format, Ruff lint, and ty checks passed for all changed files.


strickvl (dismissed):

AlexejPenner:
image
Here's what it looks like now - so much better imo

schustmi (approved):

PR #785, opened 2026-08-14
Support read replica database endpoints

schustmi:
Route the database sessions of endpoints marked with a new `read_only` decorator to a read replica configured via `KITARU_SERVER_READ_DATABASE_URL` or `KITARU_SERVER_DB_READ_HOST`, with replica connections enforced read-only and auth writes kept on the writer engine.


schustmi:
cc @safoinme

AlexejPenner (approved):
🦭

PR #786, opened 2026-08-14
Improve CLI connection error messages

strickvl:
## Summary

CLI connection failures now identify the server that could not be reached. Login also explains when a URL does not expose a Kitaru API instead of returning the generic `Not found` message.

The transport error path preserves a selected base path when the failed request uses the same host. If a secondary service such as the control plane fails, it reports that failed request origin instead. Request paths, query strings, fragments, and URL credentials are not included in fallback output.

### Reproduction

Login against a URL without a Kitaru API now reports:

```console
$ uv run kitaru login https://preview.example.com
Error: Kitaru is not available at https://preview.example.com. Check the URL or deployment.
```

A worker connection failure now reports the selected server:

```console
$ uv run kitaru worker start --name returns-example-worker
starting: {"claim_batch_size": null, "concurrency": 1, "job_id": null, "kinds": null, "name": "returns-example-worker", "selectors": null}
Error: The server at https://api.example.com is unavailable: All connection attempts failed
Hint: Check the selected server with `kitaru status`, then retry.
```

Please pay particular attention to the origin-aware transport mapping in `src/kitaru/cli/app.py`. The full selected URL is used only when it matches the origin of the failed request.

The latest `develop` OpenAPI change had not regenerated `packages/core/src/generated/openapi.ts`. The first CI run exposed that mismatch, so this PR also includes the generated TypeScript sync required by the repository contract.

### Validation

- `just check`
- `uv run pytest tests/cli -q` (`369 passed`)
- TypeScript generate-check, lint, typecheck, build, package smoke, and `313` tests
- GitHub CI (`39/39` checks passed)
- Manual login check against the reported staging URL
- Manual worker check against an unavailable local port

## Post-deploy monitoring

No additional operational monitoring is required. During the next release smoke test, confirm that login 404s name the attempted URL and worker connection failures name the selected server without request paths or query values. Revert this commit if the CLI reports the wrong host or exposes request-specific URL data.

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/packages/core/src/generated/openapi.ts
+++ b/packages/core/src/generated/openapi.ts
@@ -1900,6 +1900,7 @@ export interface paths {
          *
          *     Args:
          *         settings: Service settings governing auth behavior.
+         *         server_id: Persisted server id.
          *         ui_version: UI version served by this process.
          *
          *     Returns:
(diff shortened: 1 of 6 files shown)


AlexejPenner (dismissed):
🦭

AlexejPenner (approved):
🦭

PR #791, opened 2026-08-16
Point Kitaru docs to the public template

strickvl:
## What changed?

- Make the public [`zenml-io/kitaru-template`](https://github.com/zenml-io/kitaru-template) repository the canonical maintained returns implementation.
- Give the template README sole responsibility for clone, frozen installation, workspace setup, agent registration, worker startup, and checked-in trace import.
- Start Kitaru's returns tutorial from the verified `returns-resolver@1` version and ten imported `returns-baseline` sessions, then keep the tutorial focused on Observe, Judge, Define, Replay, and Compare.
- Preserve the generic evaluator example in the tutorial and align the documented worker name with the template.
- Update repository entry points and contributor guidance to use the same template-to-tutorial handoff.

## Why?

The returns implementation now has a public home in `zenml-io/kitaru-template`. Keeping setup and import instructions there avoids maintaining two copies of the same walkthrough while letting Kitaru's docs explain the investigation method in depth.

This PR documents the new boundary. Removal of the remaining in-tree implementation and cross-repository compatibility enforcement remain part of the later cutover.

Related: [zenml-io/zenml-frontend-monorepo#321](https://github.com/zenml-io/zenml-frontend-monorepo/issues/321)

## Reviewer Notes

The important handoff begins in the public template's root README and continues at `docs/book/tutorials/returns-agent/README.md`. The tutorial no longer repeats registration or import commands. Its Observe page verifies the objects created by the template, while Define retains the evaluator guidance that would otherwise disappear when the in-tree example is removed.

@AlexejPenner, please confirm the public-facing wording and the template-to-tutorial handoff in particular. This should also stay aligned with the frontend copy tracked in issue 321.

## Reproduction

1. Open the [`kitaru-template` setup](https://github.com/zenml-io/kitaru-template#prepare-the-template) and confirm it owns the path from cloning through the ten-session check.
2. Open the [returns-agent tutorial on this branch](https://github.com/zenml-io/kitaru/blob/feat/template-repository-docs/docs/book/tutorials/returns-agent/README.md) and confirm its prerequisites are exactly `returns-resolver@1`, ten `returns-baseline` sessions, and the running template worker.
3. Continue to [Observe](https://github.com/zenml-io/kitaru/blob/feat/template-repository-docs/docs/book/tutorials/returns-agent/observe.md) and confirm it does not ask the reader to register or import the same data again.
4. Continue to [Define](https://github.com/zenml-io/kitaru/blob/feat/template-repository-docs/docs/book/tutorials/returns-agent/define.md) and confirm the generic evaluator example remains available without depending on the in-tree example README.

## Local checks run

- `just check`
- `just links-external` (401 links checked, 0 errors)
- `just canonical-example-test` (8 passed)


AlexejPenner (approved):
🦭

Issue #794, opened 2026-08-17
Default `dashboard_url` to the server URL for bundled-UI deployments, expose review links in CLI output


strickvl:
## Problem

Originally this issue asked for a server-computed `review_url` because clients had no reliable base URL to build review links from. Most of that is resolved: #796 points the managed `dashboard_url` at the Kitaru workspace base (`/kitaru-workspaces/WORKSPACE_ID`), so appending the stable frontend routes now works. Device auth is unaffected: managed login gets its verification URL from the control plane, not from `dashboard_url`.

What remains is the OSS gap. A server that hosts the bundled UI itself still reports `dashboard_url: null` unless someone configures it, even though `ui_version` shows a UI is being served. So the default local setup is the one case where clients still can't build a link.

## Remaining work

1. When the server hosts the bundled UI and no dashboard URL is configured, default `dashboard_url` to the server's own base URL instead of `null`.
2. Have `kitaru investigation create` put the review link in the existing top-level `links` output as `links.review`, and show it in human output. MCP gets the same data through the SDK response.

A server-computed `review_url` field on `InvestigationResponse` is no longer required for this to work; it would just save clients the route appending. Can be split out or dropped.

schustmi:
@strickvl The server has exactly the same amount of information about frontend routing as the CLI, so I don't see a difference tbh.

strickvl:
Rescoped after the discussion with @schustmi and #796. The managed base URL is fixed there, and device auth turned out not to depend on `dashboard_url` on managed deployments (the control-plane login flow supplies its own verification URL). What's left here: the OSS bundled-UI default and the CLI `links.review` promotion.

strickvl:
Closing after discussion. #796 fixed the managed base URL, and the OSS case resolves client-side the way ZenML's `dashboard_utils` does it: if the server reports no `dashboard_url` but serves a bundled UI (`ui_version` set), the client uses the server URL it logged into as the UI base. No server changes needed. The remaining CLI work (resolve the review link with that rule and expose it as `links.review`) is small and will go up as a PR directly.

PR #798, opened 2026-08-17
Resolve the dashboard review link in `kitaru investigation create`

strickvl:
Closes the client-side half of the discussion in #794. After #796 fixed the managed dashboard URL, the agreed approach is that clients resolve the review page themselves, the way ZenML's `dashboard_utils` does: use the server-stated `dashboard_url` when there is one, and for servers hosting the bundled UI fall back to the URL the client logged into (`ui_version` in `/api/v1/info` proves a UI is being served). A server behind a proxy cannot state its own public address, so the login URL is the only trustworthy base in that case.

## What changed

- New `kitaru.client.dashboard_urls` module with `get_dashboard_base_url` and `get_investigation_review_url`. Pure functions over `ServerInfoResponse`; resolution order is `dashboard_url` → (`server_url` or the client's API URL) when `ui_version` is set → `None`.
- `kitaru investigation create` fetches `/api/v1/info` after creating and puts the resolved URL in the existing top-level `links` output as `links.review`. Human output renders it through the existing footer; `--output json` carries it as `links.review`.
- The helper lives in the client package (not the CLI) so MCP, which creates investigations through the same SDK, can reuse it later.

## Reviewer Notes

The risky part is the fallback rule in `get_dashboard_base_url`. If it is wrong in the "no dashboard_url, no ui_version" case, an API-only deployment would get a confident link to a page that does not exist — so that case deliberately returns `None` and the CLI omits the link without a warning (a UI-less deployment is a valid state, not an error). If it is wrong in the bundled-UI case, local onboarding hands out a broken link; the tests in `tests/client/test_dashboard_urls.py` pin the precedence including trailing slashes.

The second thing worth attention is failure behavior in `create_investigation`: the info request is decoration, so an `APIError`, a transport error, or a malformed info payload from a version-skewed server must not fail a create that already succeeded. It downgrades to a warning on stderr and the investigation is still returned. `tests/cli/test_investigations.py::test_create_warns_when_server_info_is_unavailable` covers this.

The route literal `/agents/{agent}/investigations/{id}/review` is the stable shape both frontends serve (direct UI at its root, managed under the `/kitaru-workspaces/{ws}` base that `dashboard_url` now points at after #796).

### Reproduction

Against a local server with the bundled UI:

```bash
kitaru investigation create triage --agent AGENT --session SESSION --session-question "SESSION:root-cause=What caused it?" --output json
```

Expect a top-level `links.review` of the form `http://localhost:8000/agents/AGENT_ID/investigations/INV_ID/review`, and the same URL as a labeled `review:` line in human output. Against a managed workspace (post-#796 Helm value) the link starts with the `/kitaru-workspaces/WORKSPACE_ID` base. Against an API-only server there is no `links.review` and no warning.

Local checks run: `just fix`, format/lint/typecheck/typos/yaml/openapi/actions checks, and the full test suite (10 pre-existing failures reproduce identically on clean `origin/develop`: `tests/server/test_analytics_source.py`, `test_deadlock_handler.py`, `test_pool_timeout_handler.py`, `tests/typescript/test_mastra_adapter.py`; unrelated to this change). `/simplify` was run and applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
+## [Unreleased]
+
+### Added
+
+- `kitaru investigation create` returns the dashboard review page as a `review` link, resolved from the server-stated dashboard URL or, for servers hosting the bundled UI, from the URL the client logged into.
+
 ## [0.22.0rc7]
 
 ### Changed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,12 +5,6 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/),
 and this project adheres to [Semantic Versioning](https://semver.org/).
 
-## [Unreleased]
-
-### Added
-
-- `kitaru investigation create` returns the dashboard review page as a `review` link, resolved from the server-stated dashboard URL or, for servers hosting the bundled UI, from the URL the client logged into.
-
 ## [0.22.0rc7]
 
 ### Changed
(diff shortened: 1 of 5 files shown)


schustmi (dismissed):

strickvl:
Addressed both review comments in d1b56d83: the info-fetch except now also catches `ValueError` (covers JSON decode and Pydantic validation errors from a malformed or version-skewed `/api/v1/info` payload), with the warning test parametrized over both failure kinds, and the duplicate `[Unreleased]` changelog section is merged into the existing one.

PR #803, opened 2026-08-17
Pre release fixes

schustmi:
- Merge the `002_task_claim_indexes` migration into `001_initial` so the pre-release schema ships as a single revision
- Send the skill named in `KITARU_ACTIVE_SKILL` as an `X-Kitaru-Skill` header from the Python and TypeScript clients and stamp it onto server-side analytics events
- Identify the MCP server and CLI as `kitaru-mcp` and `kitaru-cli` in the client identification header instead of `kitaru-python`
- Reuse stored control plane credentials on `kitaru login` instead of re-running the device flow every time, with `--refresh` forcing a new device login
- Raise a clear client error when a server answers an API request with an HTML page, instead of surfacing a raw JSON decode failure
- Add `Account Created` and `Worker Registered` analytics events, and name the plugin and provider on the import and evaluation events
- Serve the UI externally on pro servers by setting `KITARU_SERVER_EXTERNAL_UI` in the helm chart when `server.pro.enabled` is set


AlexejPenner (approved):
:seal:

PR #805, opened 2026-08-17
Fix the `.mcp.json` snippet for `kitaru-mcp` so it can actually start

strickvl:
The `.mcp.json` snippet on the "Drive it from your coding agent" page could not work as printed, so anyone copy-pasting it to wire Kitaru into Claude Code, Cursor, or a similar assistant hit a dead end before they saw a single tool. A third problem showed up in the wording around it: the example URL invited people to treat `localhost` as *the* Kitaru server rather than as one possible target.

Three separate failures, all fixed here:

1. **`command: "kitaru-mcp"`.** Kitaru is normally installed with uv into a project virtual environment, which puts the `kitaru-mcp` console script in `.venv/bin/`. The assistant starts an MCP server as a plain subprocess and does not activate any environment, so the bare name is not on `PATH` and the launch dies with `ENOENT: Executable not found in $PATH: "kitaru-mcp"`.
2. **`args: []`.** With no `--server`, no `KITARU_MCP_SERVER`, and no `KITARU_API_URL`, the server has no target and exits at startup with `No Kitaru server URL is configured for the MCP process.`; the host surfaces that as `MCP error -32000: Connection closed`. The paragraph immediately below the snippet already said startup fails when nothing selects a server, so the page contradicted its own example.
3. **Nothing said the `--server` URL has to match the server you are logged into.** The MCP server resolves its target once at startup and does not follow the CLI's current selection. Point it at the wrong URL and nothing complains: the assistant reports the Kitaru server as connected, and every tool call reads from a server the user never logged into, which usually presents as a mysteriously empty workspace. That is a worse outcome than the two hard failures above, because there is no error message to search for.

The snippet now uses `uv` as the command (`uv` is on `PATH`, so this avoids both the virtual-environment problem and baking a machine-specific absolute path into published docs), passes an explicit `--server http://localhost:8000`, and passes `--mode standard`. The prose after the block explains each one, and is explicit that the URL is a placeholder: `http://localhost:8000` is only correct after `kitaru login --local`, managed and self-hosted users substitute their workspace URL, and `kitaru status` is how you find out which server you are actually on: its Connection section prints the server URL it resolved, and Health reports `Authentication: authenticated` when your stored credential works against that server.

`--mode standard` is deliberate rather than incidental. Tools above the current capability mode are never registered, so on the default `read-only` an assistant trying to work the investigation loop sees no write tools at all and silently falls back to shelling out to the CLI, which is exactly the behavior the MCP server exists to replace. The page still recommends starting `read-only`; it now says what to expect if you do, and points forward to the capability-mode table rather than restating it.

The install line also moves from `pip install "kitaru[mcp]"` to `uv add "kitaru[mcp]"`, matching how every other page in `docs/book/` phrases installs.

## Reviewer Notes

This is a small diff with three specific failure modes behind it, so the thing worth checking is not the wording but whether the snippet actually starts a server and talks to the right one.

The whole change lives in `docs/book/agent-native/mcp-server.md`: the fenced `json` block, three short explanatory paragraphs replacing nothing that was there before, and the `uv add` line. Nothing else on the page moved. The tool/mode table, the capability-mode discussion, the target-resolution paragraph, and the guardrails section are untouched.

What breaks if this is wrong: this snippet is the first thing a new user copies, and a failing MCP launch looks like a Kitaru bug rather than a docs bug, because the assistant reports only "connection closed" with no explanation. The target-mismatch case is subtler and worth reading closely, since a reviewer is the last line of defense on it. If the page ever goes back to presenting `localhost` as the answer instead of as an example, a user on a managed workspace gets a connected-looking server wired to a different backend, and the first symptom they see is that Kitaru appears to have no sessions in it.

The claims about `kitaru status` are checkable against `src/kitaru/cli/diagnostics.py` and `src/kitaru/cli/presentation.py`, and this is the one place worth being pedantic. The status result carries three distinct fields: `server_url`, `server_source`, and `credential_status` (the last from `_credential_summary`, which yields `stored` / `environment` / `none`). In human output, `presentation.py` renders `server_url` as **Server** and `server_source` as **Source** under Connection, so the `Source: stored` a user sees reports where the *server URL* came from, not where the credential came from. An earlier draft of this page described that line as a credential source; the wording now points at `Authentication` in the Health section instead, which is the field that actually confirms the login worked. `credential_status` is available in `--output json`.

One judgment call worth a reviewer's eye: keeping `http://localhost:8000` in the snippet at all. The alternative is showing `KITARU_MCP_SERVER` in an `env` block, or a `<your-server-url>` placeholder that cannot be pasted as-is. I kept the concrete localhost value because it is genuinely correct for the first-run path (`kitaru login --local`) and a runnable snippet beats an illustrative one, then made the substitution instruction the most prominent sentence in the surrounding prose.

### Reproduction

Prove the old snippet fails, in a project where Kitaru was installed with uv:

```bash
uv add "kitaru[mcp]"

# 1. The old command, exactly as the page printed it:
env -i PATH="/usr/bin:/bin" kitaru-mcp
# -> "command not found: kitaru-mcp"  (this is what the assistant sees)

# 2. Even with the binary resolvable, the old empty args fail:
env -u KITARU_API_URL -u KITARU_MCP_SERVER uv run kitaru-mcp
# -> "No Kitaru server URL is configured for the MCP process." and exit
```

Now prove the new one works:

```bash
kitaru login --local          # provisions the server at http://localhost:8000
kitaru status                 # confirm the Server URL and Authentication: authenticated

# Speak MCP at it directly and check the handshake plus the tool list:
printf '%s\n%s\n%s\n' \
  '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"repro","version":"0"}}}' \
  '{"jsonrpc":"2.0","method":"notifications/initialized"}' \
  '{"jsonrpc":"2.0","id":2,"method":"tools/list"}' \
  | uv run kitaru-mcp --server http://localhost:8000 --mode standard
```

Look for a successful `initialize` result followed by a `tools/list` response containing the `standard` tools (`kitaru_cohorts_manage`, `kitaru_review_manage`, `kitaru_workflow_start`, and friends) alongside the read-only ones. Then drop `--mode standard` and run it again: the same handshake succeeds but only the three `*_read` tools come back, which is the silent-fallback case the new prose warns about.

To see the target mismatch the new paragraph is about, log in to a second server, then start the MCP process against the *first* URL and list sessions through it. The handshake still succeeds and the tools are all present; the results just come from the other server.

The quickest tell in that situation is the server's own startup line on stderr, which is worth knowing about while reviewing this: it logs `target=<url> credential_source=stored` when the URL matches a saved login, and `credential_source=none` when it does not. That line, not the host's connected/failed indicator, is what tells you whether the process can authenticate: "connected" only means the subprocess answered the MCP handshake.

End to end, paste the new block into a project's `.mcp.json`, restart Claude Code, and run `/mcp`: the `kitaru` server should report connected with its tools listed.

**Local checks run:** `just check` passes on format, lint, OpenAPI, typecheck, yaml, actionlint, and the offline `links` job (339 links OK, 0 errors). The `typos` stage fails in my worktree, but only on unrelated untracked trace fixtures sitting in that directory; `uvx typos docs/` is clean. No `CHANGELOG.md` entry: that file currently has no `[Unreleased]` heading (the top section is a release candidate), so there is nowhere for a docs fix to land without inventing one.


AlexejPenner (dismissed):
🦭

PR #813, opened 2026-08-17
Pre-release server hardening from stress testing

schustmi:
Hardening fixes surfaced by stress-testing the server against concurrent load, malformed input, and oversized payloads.

- Liveness-check pooled database connections on checkout so a connection dropped while idle is replaced, not failed.
- Return 503 instead of a raw 500 when the database connection is lost or the database is recovering.
- Reject negative session node indexes.
- Reject annotation spans with negative offsets or an end before the start.
- Reject over-deep filter expressions with a validation error instead of a RecursionError.
- Return 422 instead of a raw 500 when a run spec references a secret id that does not exist.
- Cap the session node batch at 500 to stay under PostgreSQL's bind-parameter limit.


AlexejPenner (dismissed):
:seal:

AlexejPenner (approved):
:seal:

PR #835, opened 2026-08-18
Entity deletion

### LINKED ISSUE #850: Implement correct deletion behaviour

schustmi:
This PR gives every entity an explicit deletion rule. Deleting an agent hides it and retains its subtree, a delete that would orphan a referenced row is rejected with 409, and everything else cascades or nulls a pointer according to one table that a test enforces.


src/kitaru/server/application/services/experiment_service.py
```diff
@@ -339,6 +350,16 @@ async def delete_experiment(
         """
         _ = actor
         experiment = await self._repository.get(experiment_id)
+        runs = await self._experiment_runs.list_by_experiment(experiment_id)
```
schustmi, line 359: Concurrent experiment run creation will probably mess with this

src/kitaru/server/application/services/task_service.py
```diff
@@ -367,7 +382,6 @@ async def propagate_job_cancel(self, job_id: uuid.UUID) -> None:
             DBAPIError: Another transaction holds one of the task rows.
         """
         await self._transitions.request_jobs_cancel([job_id], nowait=True)
```
schustmi, line 384: Why don't we need to settle anymore?

src/kitaru/server/adapters/db/orm/task.py
```diff
@@ -53,12 +53,17 @@
 ON_FAILURE_LENGTH = 16
 
 TASK_JOB_ID_FOREIGN_KEY = foreign_key_name("task", ["job_id"])
+TASK_WORKER_ID_FOREIGN_KEY = foreign_key_name("task", ["worker_id"])
+# A task names the rows it takes as input by id alone. The queue holds work
+# that outlives the rows it points at, so deleting one of them neither cascades
+# into the queue nor is blocked by it, and a claim that cannot resolve an input
+# cancels the task instead. The names stay here for the migration that drops
+# the constraints.
 TASK_AGENT_VERSION_ID_FOREIGN_KEY = foreign_key_name("task", ["agent_version_id"])
```
schustmi: These can be removed I think?

AlexejPenner (approved):
🦭

PR #837, opened 2026-08-18
Unify all examples under `examples/python/` and `examples/typescript/`

strickvl:
## What and why

The v1 to v2 transition left three example locations: `examples/` (Python adapter examples under `integrations/`), `examples/v2/` (a single MCP config), and `v2_examples/` (the TypeScript examples). The split was never a version split: it was Python vs TypeScript mislabeled as old vs new, and it had already produced a regression test whose only job was to stop the coverage audit from confusing the two prefixes.

This PR unifies everything under one tree:

```
examples/
  python/      langgraph_v2, openai_agents_v2
  typescript/  mastra_support_triage, vercel_ai_support_triage, vercel_ai_ticket_resolver
```

`examples/v2/mcp` is deleted rather than moved: it was two files (a README and a `.mcp.json.example`) duplicating `docs/book/agent-native/setup.md`, which already covers the same setup in more detail and stays the single source.

## Reviewer Notes

The riskiest part of a folder move like this is not the rename itself but the paths that count directory levels. The TypeScript examples went from two levels deep (`v2_examples/x`) to three (`examples/typescript/x`), so every repo-root-relative reference inside them needed one more step up: `Path(__file__).resolve().parents[N]` in the Python drivers, the `extends: ../../tsconfig.base.json` lines, the `link:../../packages/*` dependencies in the standalone ticket resolver, and the `node ../../scripts/clean-typescript-dist.mjs` build steps. If one of these were wrong, the failure would be a runtime path error rather than a compile error, so the e2e run below is the meaningful proof.

Two strings are semantic, not cosmetic: the `RUN_COMMAND` constants in the Mastra and Vercel AI support-triage demos are stored in Kitaru as the agent version's run command, so their update changes recorded data for new runs.

The coverage machinery got simpler because the unification obsoleted parts of it: `scripts/audit-example-coverage.py` loses the `v2_examples` special cases and the `explicit_exclusions` parsing (its only consumer was the deleted MCP entry), and the prefix-disambiguation test in `tests/test_example_coverage_audit.py` is repurposed to pin URL path extraction for the new nested layout.

Three adjacent fixes ride along, each in its own commit:

- The `just links` recipes passed their own `--exclude-path` flags, which override rather than merge with `lychee.toml`, so local link checks and the CI links job scanned different file sets. The exclusion list now lives only in `lychee.toml`.
- The pool-timeout handler tests failed on any machine with a downloaded UI bundle, because the bundled UI catch-all route shadowed their probe routes (CI never sees this since a fresh checkout has no bundle). They now build the app through the same no-bundle seam `test_ui_serving.py` uses. A deeper fix would be a `SERVE_UI` setting on `APISettings`; left out of scope here since three server tests currently work around this in three different ways and deserve one deliberate change.
- The ty exclude narrows from `examples/**` to `examples/python/**` (those import plugin-workspace packages absent from the root environment), so the TypeScript examples' Python drivers are now type-checked for the first time.

External surfaces: the GitHub `tree/develop/...` links in the adapter docs and the three npm package READMEs are updated in the same change, so the GitBook pages and the next npm publishes point at the new paths. Old deep links into `v2_examples/` on GitHub will 404 after merge; nothing in `kitaru-template` or `kitaru-skills` references these paths (verified), and the one stale cross-repo link found in the zenml docs predates this change and is tracked in zenml-io/zenml#5160. The new `tree/develop/examples/...` links 404 until this PR merges, so `lychee.toml` carries a temporary exclusion for them (same pattern as the existing v2 docs one); a follow-up should remove both once merged and live.

### Reproduction

The full end-to-end proof from the new location (requires Node 22 and the Postgres container):

```bash
docker compose up -d db
cd examples/typescript/vercel_ai_ticket_resolver
pnpm --ignore-workspace install --frozen-lockfile
pnpm build && pnpm test && pnpm test:evaluator && pnpm typecheck && pnpm lint && pnpm test:e2e
```

`test:e2e` boots the real API against Postgres, records the TypeScript agent, runs the evaluator, builds a cohort, and replays it; a wrong depth-relative path anywhere in the moved tree fails this run. For the Python side:

```bash
uv sync --project plugins --frozen --all-packages
uv run --project plugins python -m examples.python.langgraph_v2
just example-coverage-audit
```

Local checks run: `just check`, `just test` (full suite, 3577 passed), root `pnpm run build/lint/typecheck/test:built`, both pnpm lockfiles regenerated, offline `just links` clean.


htahir1 (dismissed):

AlexejPenner:
@strickvl we should talk about the purpose of examples vs the template repo

AlexejPenner (dismissed):
🦭

PR #838, opened 2026-08-18
Enable Ruff pydocstyle (`D`) rules for Google-style docstrings

### LINKED ISSUE #809: Enable Ruff checks for Google-style docstrings
## Summary

Enable Ruff's pydocstyle rules for the Python codebase so the configured Google-style docstring convention is actually enforced in CI.

cc @schustmi

## Current state

`pyproject.toml` sets:

```toml
[tool.ruff.lint.pydocstyle]
convention = "google"
```

But the active Ruff rule list does not include `D`, so the normal `ruff check .` run does not check docstrings.

An explicit audit with `ruff check --select D .` currently reports 228 docstring-related violations, including:

- 23 undocumented public methods (`D102`) in scripts and tests
- missing public class and magic-method docstrings
- Google-style formatting issues such as missing summary/description separation and multiline closing-quote placement

The shipped package under `src/kitaru` has no concrete public-method `D102` violations, but it still has a small number of class and magic-method findings.

## Proposed work

- Add the appropriate Ruff `D` rules to the active lint configuration.
- Decide whether enforcement should cover `src/kitaru`, `scripts`, and `tests` uniformly.
- Fix the existing baseline violations, or add narrow, documented per-file ignores where test doubles or magic methods are intentionally exempt.
- Ensure the standard CI lint command catches future missing or malformed Google-style docstrings.

## Acceptance criteria

- `ruff check .` enforces the agreed docstring rules.
- The Google-style convention remains the source of truth.
- CI passes with no unexplained docstring violations.
- Any exclusions are explicit, narrow, and documented.

strickvl:
Closes #809.

## What this does

`pyproject.toml` has declared `convention = "google"` under `[tool.ruff.lint.pydocstyle]` for a while, but the `D` rule family was never in the `select` list, so the convention was configured yet unenforced — `ruff check .` (and therefore `just check` and CI lint) passed regardless of docstring state. When the issue was filed the audit showed 228 violations; by the time of this branch it had grown to 310, which is the problem in miniature: without the rule enabled, the backlog only grows.

This PR enables `D` and clears the baseline:

- **Enforcement scope:** full `D` enforcement on `src/` and `scripts/`. `tests/**` gets a per-file-ignore for the missing-docstring rules only (`D1`), so test helpers may omit docstrings, but any docstring a test *does* have must still satisfy the Google formatting rules (`D2xx` and up). `scripts/google_adk_fakes.py` also skips `D1`: it is a file of test doubles whose methods only mirror the real ADK surface, and the class-level docstrings already say what each fake stands in for. Both exemptions are commented in `pyproject.toml`.
- **Baseline fixes:** 24 auto-fixed `D209` closing-quote placements, 24 `D205` summary/description reflows in `tests/`, missing docstrings added across `scripts/` and `src/kitaru/mcp/`.
- **MCP schema snapshots regenerated:** two of the new docstrings sit on Pydantic models in `src/kitaru/mcp/models/activity.py`, and Pydantic publishes a model's docstring as the `description` field of its JSON schema, so the committed MCP tool-schema snapshots changed.
- **Lychee exclusion widened:** the offline link check previously excluded only `docs/node_modules`, so running the TypeScript tests locally (which installs `v2_examples/*/node_modules`) made a subsequent `just check` fail on third-party package READMEs. The recipe now excludes all `node_modules` directories; the scanned-file count for tracked docs is unchanged (515 before and after).

The `plugins/` tree is intentionally untouched — it is already excluded from this Ruff configuration and keeps its own lint setup.

## Reviewer Notes

The riskiest part of this PR is not the docstrings, it is the two places where a docstring is *not* just documentation:

1. **`src/kitaru/mcp/models/activity.py` + `tests/mcp/snapshots/*.json`** — the docstrings on `SessionNodesRequest` and `SortedChildrenRequest` become part of the published MCP tool schemas. The snapshot diff is small (two `description` fields plus updated byte counts in `metrics.json`), but it is real API surface. The wording was deliberately kept short because `kitaru_activity_read` sits close to its per-tool schema-size budget in `tests/mcp/test_registry_runtime.py`; after this change it is 33661 bytes against a 33792-byte budget. If you would rather keep these models out of the schema entirely, the alternative is dropping the two docstrings and their snapshot changes.
2. **`pyproject.toml`** — the shape of the per-file-ignores decides what CI enforces forever after. `"tests/**" = ["D1"]` is prefix-granular on purpose: listing `D100`–`D107` individually would silently miss future rules in that family.

Everything else is mechanical: the `tests/` changes only reshape existing docstrings (summary line, blank line, description — no test logic touched), and the `scripts/`/`src/kitaru/mcp/` changes only add docstrings. The `Justfile` change affects the local `links`/`links-external` recipes only, not what CI installs.

### Reproduction

```bash
# On develop: the convention is configured but nothing enforces it
git checkout develop && uv run ruff check --select D . | tail -1   # ~310 errors
uv run ruff check . | tail -1                                       # passes anyway

# On this branch: the same command now enforces docstrings
git checkout feat/enable-pydocstyle && uv run ruff check . | tail -1  # passes, D active
# Delete any docstring in src/kitaru/ and rerun — ruff now fails with a D1xx error.
```

Local checks run: `just check` and `just test` (3577 passed, 16 skipped) both green, including the MCP schema-budget tests after snapshot regeneration.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -72,11 +72,11 @@ audit:
 # Source MDX uses docs-app-root routes such as /guides/...; site-build validates
 # those after materializing public /docs/... links, where lychee can resolve them.
 links:
-    lychee --offline --root-dir . --exclude-path '.venv' --exclude-path 'docs/node_modules' --exclude-path 'docs/out' --exclude-path 'design' './**/*.md'
+    lychee --offline --root-dir . --exclude-path '.venv' --exclude-path 'node_modules' --exclude-path 'docs/out' --exclude-path 'design' './**/*.md'
 
 # Check raw Markdown links including external URLs (slow, used in CI)
 links-external:
-    lychee --root-dir . --exclude-path '.venv' --exclude-path 'docs/node_modules' --exclude-path 'docs/out' --exclude-path 'design' './**/*.md'
+    lychee --root-dir . --exclude-path '.venv' --exclude-path 'node_modules' --exclude-path 'docs/out' --exclude-path 'design' './**/*.md'
 
 # Auto-fix formatting, lint issues, and YAML
 fix:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -142,7 +142,16 @@ select = [
     "B",    # bugbear
     "SIM",  # simplify
     "RUF",  # ruff-specific rules
+    "D",    # pydocstyle (Google convention, see [tool.ruff.lint.pydocstyle])
 ]
+
+[tool.ruff.lint.per-file-ignores]
+# Tests may omit docstrings (D1xx), but any docstring they do have must still
+# follow the Google formatting rules (D2xx and up).
+"tests/**" = ["D1"]
+# Test doubles whose methods only mirror the real ADK surface; class-level
+# docstrings already say what each fake stands in for.
+"scripts/google_adk_fakes.py" = ["D1"]
 
 [tool.ruff.lint.flake8-bugbear]
 extend-immutable-calls = ["fastapi.Depends"]
--- a/scripts/audit-example-coverage.py
+++ b/scripts/audit-example-coverage.py
@@ -53,6 +53,7 @@ SHELL_PLACEHOLDER_ASSIGNMENT_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*=<[^<>\s]+
 
 
 def main() -> int:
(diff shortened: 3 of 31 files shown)


schustmi (approved):

AlexejPenner (approved):
🦭

PR #842, opened 2026-08-19
Register a new worker on every start

schustmi:
Worker names are labels now, every start registers a new worker, tokens renew through POST /api/v1/workers/{worker_id}/token, and listings hide stale workers unless include_stale is set.


AlexejPenner (dismissed):
🦭

strickvl (dismissed):

strickvl (dismissed):

PR #849, opened 2026-08-20
Add Arize Phoenix trace importer

### LINKED ISSUE #853: Add Arize Phoenix trace importer
Following external request.

strickvl:
## What changed?

Kitaru can now import Arize Phoenix trace files directly with `kitaru/phoenix`. The built-in importer accepts Phoenix UI JSONL downloads and Phoenix CLI JSON trace envelopes, then produces one Kitaru session per Phoenix trace with its span graph and original evidence intact.

The new importer is included in the default plugin catalog and release inventory. The documentation explains the supported file shapes, normalization contract, identity rules, and known limits.

Companion format specimen: https://github.com/zenml-io/kitaru-trace-corpus/pull/2

@htahir1, this is the file-based Phoenix importer prompted by the recent Arize request. It does not call the Phoenix API or require Phoenix credentials.

## Why?

Phoenix users can already download the trace data Kitaru needs, but they previously had to write a custom conversion. This importer makes those files usable without changing Phoenix as their trace system of record.

The importer uses `(trace_id, span_id)` as source identity. It reconstructs parent-child relationships, maps LLM and tool spans to typed Kitaru nodes, preserves unknown attributes and events, and isolates malformed lines and traces so valid traces in the same file still import.

## Reviewer Notes

Start with `plugins/packages/phoenix-importer/src/kitaru_phoenix_importer/importer.py`. The important contract is the split between file containers and span normalization: UI exports provide flat span rows, while CLI retrieval wraps the same span objects in `{traceId, spans}` envelopes.

Session status follows a genuine null-parent root when the export contains one, not a missing-parent orphan or failed child. This keeps partial exports from changing the session name or status while preserving recovered tool failures as failed nodes inside a completed run.

The importer supports both OpenTelemetry GenAI attributes and Phoenix's indexed OpenInference `llm.*` message attributes. Structured messages take precedence over opaque `output.value` stubs. Raw Phoenix attributes, events, annotations, and notes remain attached to the imported evidence.

The adversarial review findings were reproduced against the implementation before fixes. The accepted fixes cover root selection, malformed-line isolation, UTF-8 BOMs, split-envelope metadata, failure line provenance, message fidelity, tool schemas, legacy token names, and resolved model selection. DataFrame JSONL serialization remains deliberately unsupported and is now documented.

## Reproduction

Run the focused importer tests:

```bash
uv run --project plugins pytest plugins/tests/importers/test_phoenix.py -q
```

Then build and install-test the standalone plugin artifact:

```bash
uv run --no-sync python scripts/smoke_plugin_artifacts.py \
  --package plugins/packages/phoenix-importer
```

To exercise real provider data, check out the companion corpus PR and parse the nine files under `exports/arize/**/api/spans.jsonl` and `exports/arize/**/ui/traces.jsonl`. The final validation produced 11 sessions and 84 nodes with zero import failures. The downloaded Phoenix UI file produced one session and nine nodes with zero failures.

## Local checks run

- `just test`: 3,583 passed, 16 skipped.
- Focused Phoenix importer tests: 21 passed.
- `uv run --project plugins pytest plugins/tests -q`: 411 passed.
- `just check`: format, lint, OpenAPI, type checks, typos, YAML, Actions, and links passed.
- Plugin artifact smoke test passed.
- All nine real Phoenix files parsed into 11 sessions and 84 nodes with zero failures.
- The downloaded Phoenix file parsed into one session and nine nodes with zero failures.
- Simplify used three independent finding-only reviewers plus an inline altitude pass; it removed duplicate test setup and avoided unnecessary message scans and repeated metadata copies.

- During the first importer release, run one known Phoenix UI JSONL import and one CLI-envelope import on staging.
- Search worker logs for `Phoenix trace`, `Phoenix span`, `parent cycle`, and `duplicate span ids` errors.
- Healthy signal: one imported session per source trace, stable trace IDs, expected node counts, and no unexpected `ImportFailure` records.
- Failure signal: a valid Phoenix export fails as a whole, loses its root graph, or produces missing model/token evidence.
- Rollback trigger: repeatable failures on either supported export shape. Remove or revert the default Phoenix importer release while preserving uploaded source files for diagnosis.
- Validation window and owner: the first 24 hours after plugin publication, owned by the Kitaru release maintainer.


src/kitaru/server/api/bootstrap.py
```diff
@@ -137,6 +137,15 @@ class DefaultPluginDefinition(FrozenModel):
...
     ),
+    DefaultPluginDefinition(
+        kind=PluginKind.IMPORTER,
+        name=f"{RESERVED_NAMESPACE}/phoenix",
+        description="Import Arize Phoenix JSON and JSONL trace exports.",
+        provider="phoenix",
+        entrypoint="kitaru_phoenix_importer.importer:parse",
+        requirement="kitaru-phoenix-importer==0.1.0",
+        display_version="0.1.0",
+    ),
```
AlexejPenner, line 148: @strickvl just fyi arize will only be shipped as default when we do 0.22.2 release of Kitaru - until then users will need to manually register arize pypi package as  valid importer

AlexejPenner (dismissed):
🦭

src/kitaru/server/api/bootstrap.py
strickvl: Can't we do an 0.22.2 release? There are various other things fixed etc since our last release?

AlexejPenner: @schustmi ?

schustmi: Sure I think so

strickvl: > arize will only be shipped as default when we do 0.22.2

Documented and settled: the guide now states that `kitaru/phoenix` is built in from Kitaru 0.22.2, and the release can follow this PR.

AlexejPenner (approved):
🦭

PR #855, opened 2026-08-20
Devtools

schustmi:
Adds `devtools/`, the shared local harness for running and testing Kitaru end to end. It was previously untracked local tooling.

- `stack.py` starts a server, database, and workers, and manages test databases including one-second snapshots.
- `traces.py` and `agent.py` generate deterministic sessions from one simulation core, so imports, live recordings, and replays line up.
- `seed.py` runs the whole flow in one command: import, live runs, cohort, experiment with a replay override, evaluations, and a check summary.
- `fixtures.py` registers the dummy agent, importer, and evaluators, and marks the onboarding survey finished so the dashboard opens on seeded data.
- `load.py`, `dbstats.py`, and `overview.py` cover concurrent load with latency percentiles, Postgres and query statistics, and a read-only server dashboard.
- `resilience.py` verifies task recovery when workers crash, hang, or die as a fleet, each scenario on its own ephemeral stack.
- Every stage can inject failures deterministically, so a given configuration reproduces the same failures.

`devtools/AGENTS.md` carries the rules and commands, and the root `AGENTS.md` points at it for changes that need a live stack.


devtools/README.md
strickvl, line 1: can this be in AGENTS.md and also can we update the AGENTS.md file in the repo base to reference this?

schustmi: Yes I wanted your opinion what's best. A skill? AGENTS.md? Any of the two but a reference to this?

strickvl: - rename the file to AGENTS.md
- reference it in the /AGENTS.md main guidance
- update the `kitaru-tests-release` skills (there are two copies) and make them aware of this

I think maybe that should be enough? but in general AGENTS.md in the root is where you should add instructions on where / how this is meant to be used.

schustmi: Tbh, the way I use this is only through manual invocations. Overkill for most things I think, every agent knows the basics themselves easily. This is more "Spawn 1000 workers with different failure modes (sigterm, sigkill) and figure out any race conditions" level stuff. Not sure that changes things?

strickvl: Yeah it's mostly just leave enough breadcrumbs that if it needs it, it can find it, but I think you probably covered most of the bases here

strickvl (approved):

PR #860, opened 2026-08-21
Clear stale entries from `pip-audit-ignored.txt` and make empty-list audits robust

strickvl:
## What

Empties the pip-audit suppression list (`.github/pip-audit-ignored.txt`) and hardens the `just audit` recipe plus the CI pip-audit step so they still run when the list is empty.

## Why

Every suppressed advisory was justified by version pins in the `zenml` dependency: `click<=8.2.1` (PYSEC-2026-2132) and `fastapi<0.121` → `starlette<0.50` (the five starlette advisories). `zenml` is no longer a dependency, so those ceilings are gone: the lockfile now resolves `click 8.4.2` and `starlette 1.3.1`, both past the fixed versions. A bare `uv run pip-audit` (no ignore flags) reports **no known vulnerabilities**, which makes all six entries stale. A stale suppression is worse than noise: if any of those IDs ever resurfaced through a downgrade, the list would silently hide it.

## Reviewer Notes

The risky part is not the deletion, it's the emptiness. Both audit entry points fed the ignore flags through `awk ... | xargs uv run pip-audit`. BSD xargs (macOS) does not run the command at all when its input is empty, so with an empty list `just audit` would exit 0 having scanned nothing. GNU xargs on the CI runners runs the command once, so CI happened to be safe — but only by xargs flavor. The Justfile recipe now uses command substitution and the CI step builds the flags with `readarray` into a quoted bash array (shellcheck-clean), so `pip-audit` always executes regardless of list length. If either of those rewrites is wrong, the failure mode is an audit that never runs — which is exactly what they exist to prevent, so please eyeball the two shell fragments.

`.github/workflows/ci.yml` is touched, so the path-filtered zizmor workflow will run on this PR.

### Reproduction

```sh
uv run pip-audit          # no ignore flags: reports "No known vulnerabilities found"
just audit                # empty list, still runs pip-audit and prints the same
```

On macOS before this change, `just audit` with the empty list printed nothing and exited 0 (BSD xargs skipped the command).

Local checks run: `just check`, `just test` (3638 passed, 16 skipped), `just zizmor` (no findings).

### DIFF
--- a/.github/pip-audit-ignored.txt
+++ b/.github/pip-audit-ignored.txt
@@ -10,26 +10,7 @@
 # To re-verify this list, run `uv run pip-audit` with no ignore flags and
 # compare the reported IDs against the entries below. Anything listed here
 # that pip-audit no longer reports is stale and should be deleted.
-
-# --- Blocked by zenml's `click<=8.2.1` cap. --------------------------------
-# zenml declares `click>=8.0.1,<=8.2.1`. The ceiling is inclusive and sits on
-# the vulnerable version itself, so no lockfile refresh can reach the 8.3.3
-# fix while we depend on zenml.
 #
-# Not exploitable here: the advisory is a command injection in `click.edit()`,
-# the helper that shells out to `$EDITOR`. Kitaru never calls it (our only
-# click usage is `click.get_app_dir()` and catching `click.ClickException`),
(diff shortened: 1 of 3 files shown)


AlexejPenner (commented):
🦭

PR #865, opened 2026-08-21
Task hooks

### LINKED ISSUE #862: Add replay worker hooks with a default copy-workdir hook for coding-agent replays
Replaying a coding-agent session (e.g. a Claude Code trace) executes inside a code repository, so a replay must not run in the worker's live working directory: it could collide with an in-progress replay or dirty the user's checkout.

Agreed scope, local workers first:

- Model pre/post replay steps as **hooks**, specified on the agent version, with some hooks built in.
- Ship a default built-in hook that copies the worker's code directory to a temp dir (including `.git` so in-repo git commands still work), runs the replay there, and deletes the copy afterwards. The copy is read-only insurance — the original checkout is never modified.
- Assumes the worker already has the correct code on its filesystem; remote code download and any Git-server connection (checkout at ref, pushing result branches) are explicitly out of scope for now and can arrive later as additional hooks.

Open question to settle during implementation: what counts as the replay's final result for a coding agent (currently the last LLM call).

schustmi:
This PR adds hooks to the agent version run spec that a worker runs around the agent process.

- `copy_workdir` runs the agent in a fresh copy of the working directory, so parallel tasks on one worker don't share state.
- `setup_command` runs a shell command in the agent's working directory before the agent process.
- `teardown_command` runs a shell command after it, with `on: "success" | "failure" | "always"` choosing the task outcome it runs on.

Hooks run in declared order before the agent process, their teardowns run in reverse order after it. A hook failure fails the task.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 - Added `DELETE /api/v1/replays/{replay_id}` to delete a standalone replay. A replay that belongs to an experiment run returns HTTP 409, since deleting the run removes its replays.
+- Added `POST /api/v1/workers/{worker_id}/token` to renew a worker token, and `kitaru worker list --include-stale` to list workers past the liveness window.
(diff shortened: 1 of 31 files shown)


AlexejPenner (approved):
🦭

PR #866, opened 2026-08-21
Move the result session link from the task to the replay table

schustmi:
Store a replay's result session id on the replay table, derive an agent task's result session through the session's task link instead of a task column, and block deleting sessions a replay references.


AlexejPenner (approved):
🦭

PR #876, opened 2026-08-24
User-defined idempotency keys

### LINKED ISSUE #667: Define retry-safe workflow creation for the v2 API and SDK
**What problem does this solve?**

Kitaru's asynchronous workflow-creation endpoints do not currently give API and SDK callers a retry-safe way to identify one logical mutation. If the server commits a workflow but the response is lost, the caller cannot know whether retrying will return the existing result or create duplicate work.

This affects:

- `POST /v1/session-runs`
- `POST /v1/replays`
- `POST /v1/evaluations`
- `POST /v1/experiments/{experiment_id}/runs`

The ambiguity is especially important for agent and MCP clients, which need to recover from dropped responses without relying on process-local memory. PR #652 deliberately omits these MCP workflow-start operations instead of inventing a CLI/MCP-local deduplication layer or changing the server and SDK from that frontend PR.

**Describe the solution you'd like**

Define a server-owned retry contract for these workflow-creation endpoints and expose it through the typed SDK. The implementation mechanism is deliberately open; the required behavior is:

- A caller can identify one logical creation request with a stable value.
- Repeating the same logical request with the same body does not create duplicate work and returns the authoritative existing response.
- Reusing the value with a different body fails with a clear conflict rather than returning unrelated work.
- The identity is scoped so different authenticated actors or routes cannot collide accidentally.
- The retry record and workflow mutation have atomic behavior: a failed or rolled-back mutation must not leave a successful-looking reservation that prevents recovery.
- The SDK exposes the supported mechanism and preserves the same value across any transport-level retry of one request.
- Mixed-version clients have a clear compatibility path, whether through server versioning, capability discovery, or a stable unsupported-response contract.

## Acceptance criteria

- The four endpoints above have documented retry behavior.
- Concurrent identical requests result in one logical mutation and equivalent authoritative responses.
- Reusing the same request identity with a changed payload returns a deterministic conflict.
- A mutation that fails before commit can be attempted again safely.
- SDK tests cover forwarding and transport retry behavior.
- Server tests cover dropped-response recovery, concurrency, payload mismatch, actor/route scoping, and rollback behavior.
- A follow-up MCP change can expose these workflow starts without local persistence or client-side deduplication.

## Out of scope

- Persisting request identities in the CLI or MCP process.
- Broadening the contract to unrelated mutation endpoints before the workflow-creation behavior is established.
- Prescribing the database schema, header name, or migration design in this issue.

Related: #624, #652

**Owner:** @schustmi · **Priority:** P1 · **Target:** to be scheduled · **Board:** Kitaru Roadmap / Backlog

schustmi:
This PR lets callers supply their own `Idempotency-Key` instead of the per-request key the transport generates, so retries of the same logical request within the retention window return the original response instead of acting twice.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 ### Added
 
 - Added `POST /api/v1/workers/{worker_id}/token` to renew a worker token, and `kitaru worker list --include-stale` to list workers past the liveness window.
+- SDK methods that call an endpoint supporting idempotency now take an `idempotency_key` argument, so callers can supply their own key instead of the per-request key the transport generates. Those endpoints declare an `Idempotency-Key` header parameter in the OpenAPI schema.
 
 ### Changed
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - Added `POST /api/v1/workers/{worker_id}/token` to renew a worker token, and `kitaru worker list --include-stale` to list workers past the liveness window.
 - SDK methods that call an endpoint supporting idempotency now take an `idempotency_key` argument, so callers can supply their own key instead of the per-request key the transport generates. Those endpoints declare an `Idempotency-Key` header parameter in the OpenAPI schema.
+- MCP tools that create a resource take an optional `idempotency_key` field, so a retried tool call whose response was lost returns the original result instead of acting twice.
 
 ### Changed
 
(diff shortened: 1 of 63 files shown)


schustmi:
@strickvl IMO this fixes #667 by allowing user-supplied idempotency keys in the SDK, CLI and MCP. The only question around this that I have: Previously, this idempotency key was meant for request retries in our api client, which sleeps for seconds before retrying a failed/timed out request. This is why the retention period for the table is 15 minutes, and any request arriving after that will not benefit from this and instead recreate the agent/session/experiment and so on.

In the agentic case, it could be that an agent shuts down and only retries it again hours or even days later, correct? How would you suggest we handle this?

strickvl (approved):

PR #879, opened 2026-08-24
Report tool call status in the replay lookup

### LINKED ISSUE #750: Distinguish recorded tool failures in the replay lookup API
## Problem

The replay lookup API cannot tell callers whether a recorded tool call completed or failed. `ToolLookupResponse` exposes only these fields:

```python
found: bool
result: Any
```

That leaves two different records with the same response:

```text
completed, outputs=null  -> {found: true, result: null}
failed, error="declined" -> {found: true, result: null}
```

The server already stores each node's status and error, but `ReplayService.tool_lookup()` returns only `node.outputs`. Adapters therefore cannot recover the missing outcome themselves.

## Expected behavior

The lookup contract should distinguish a miss, a completed hit, and a failed hit. A completed hit includes the recorded result even when it is `null`. A failed hit includes enough recorded error information for an adapter to reproduce the failure. Miss-policy behavior stays outside this endpoint and applies only when the lookup reports a genuine miss.

Scope selection is intentional:

- `scope=baseline` with an occurrence considers completed and failed tool calls in baseline order and skips in-progress calls.
- `scope=baseline` without an occurrence, `scope=agent`, and `scope=cohort_version` consider completed tool calls only.

Update the server result model and REST mapping, the public API model, OpenAPI output, and the Python and TypeScript clients. Keep the contract compatible where practical, but do not preserve the current ambiguity.

Tests should cover a completed value, a completed `null`, a failed call with a recorded error, and a miss for the same tool name, together with the scope behavior above.

## Adapter follow-up

Adapter behavior is tracked in #877 and is included in #879 following the maintainer decision to land the core and adapter changes together.

The repeated-call occurrence problem from #748 was fixed separately in #807 and is not part of this issue.

### LINKED ISSUE #877: Teach replay adapters to handle recorded tool failures
Implemented together with #750 in #879 following the maintainer decision to land the core contract and adapter behavior in one branch.

## Problem

Replay adapters currently receive only `found` and `result`. A failed tool call and a healthy tool that returned `null` therefore look the same.

The TypeScript adapters take the safe option and reject a found `null`, but that also rejects valid side-effect-only tools. Other adapters may treat the hit as a success or fall back to live execution. None of these behaviors can reproduce both recorded outcomes correctly until #750 changes the API contract.

## Expected behavior

Once #750 is merged, each adapter should follow the recorded outcome:

- For a completed call, return the recorded value, including `null`. Do not run the live tool.
- For a failed call, raise or throw a replay error at the wrapped tool-call boundary, carrying the recorded error text. Do not run the live tool or turn the failure into a regular tool result.
- For a miss, apply the configured `fail`, `error_result`, or `passthrough` behavior.

A failed hit still consumes its baseline occurrence. A miss does not. Failed calls are candidates only for occurrence-based baseline lookup; latest baseline, agent, and cohort-version lookup consider completed calls only.

Kitaru cannot recreate the original exception class from the data it stores today. That is fine for this fix: preserve failed versus completed and keep the recorded error text. The replay error aborts the adapter run unless application code catches it. Exact exception types or structured retry data would need a richer recording contract.

This covers the shared TypeScript policy used by the Vercel AI SDK and Mastra adapters, along with the PydanticAI and LangGraph adapters. Test ordinary success, successful `null`, failure, miss, and legacy-server rejection in each affected adapter. Remove the temporary fail-closed documentation and update package versions, changelogs, and core-version requirements where needed.

schustmi:
## Summary

- Replace the ambiguous `found`/`result` tool lookup response with an explicit `match` carrying the recorded status, result, and error.
- Replay completed results, including `null`, and raise stored failures without executing the live tool in LangGraph, PydanticAI, Mastra, and Vercel AI SDK adapters.
- Fail loudly when a new adapter receives a legacy lookup response without `match`, preventing an old server from being mistaken for a genuine miss and triggering passthrough side effects.
- Require Kitaru 0.22.3 for the updated Python adapters and preserve useful LangGraph error messages.

Closes #750
Closes #877

## Behavior

Occurrence-based `baseline` lookups consider completed and failed tool calls, skip in-progress calls, and consume a failed match. `baseline` lookups without an occurrence and `agent` or `cohort_version` lookups consider completed calls only, so a failed-only history is a genuine miss and follows `on_miss`.

Adapters raise a Kitaru replay error with the stored text for a failed match. They do not reconstruct the original exception class or framework retry signal, so the failure aborts the adapter run unless application code catches it.

## Reviewer Notes

The main compatibility risk is version skew. The Python response model keeps `match` optional for core-client compatibility, but adapters inspect Pydantic's field-presence metadata and reject an omitted field. The TypeScript transport validator likewise rejects an omitted field. Explicit `{"match": null}` remains a genuine miss.

Kitaru 0.22.2 does not expose the new contract. The updated Python adapters therefore require `kitaru>=0.22.3`, and the workspace development version is 0.22.3+dev so clean candidate-wheel installation exercises that dependency relationship.

### Reproduction

1. Run `uv run pytest -q tests/server/test_replays_api.py tests/server/test_session_node_repository.py` to exercise completed-null, failed, miss, and mixed occurrence behavior.
2. Run `uv run --project plugins pytest -q -c plugins/pyproject.toml plugins/tests/adapters/langgraph plugins/tests/adapters/pydantic_ai/test_adapter.py` to exercise Python adapter hits, failures, misses, and legacy-server rejection.
3. With Node 22, run `pnpm test`, `pnpm typecheck`, and `pnpm lint` to exercise the shared TypeScript policy and both TypeScript adapters.
4. Run `just plugin-artifact-smoke` to prove the Python adapter wheels install against the candidate 0.22.3 core wheel.

Full local validation also passed with `just check` and `just test` (3,735 passed, 16 skipped).

For the first coordinated release, publish Kitaru 0.22.3 before the adapter releases. For 24 hours, watch server errors on `POST /api/v1/replays/*/tool-lookup` and adapter errors containing `does not include 'match'`. Healthy behavior is an explicit `match` key on every successful lookup response and no unexpected live-tool execution after a recorded hit. If compatibility errors continue after the core rollout or a replay unexpectedly executes a side-effecting passthrough tool, pause the adapter release or pin users to the previous adapter versions while the release owner investigates.


strickvl (approved):

strickvl:
for this one @AlexejPenner when we release we have to release core before the adapters FYI

PR #888, opened 2026-08-25
Enforce `biome check` for the docs app in CI

strickvl:
## What

`docs/package.json` has had a `lint` script (`biome check`) since the docs app was created, but nothing ran it: no workflow step, no `just` recipe. On `develop` it failed with 41 errors, 13 warnings and 6 infos, so it had quietly become dead. This PR makes it pass and makes CI enforce it.

- **Apply biome's safe fixes** across the docs app: quote style, import ordering, line wrapping. This is the bulk of the diff and is mechanical. `app/global.css` only changes in quoting, hex-colour case and wrapping (verified by normalising both versions and diffing).
- **`biome.json` → `biome.jsonc`** so each rule exception carries a comment explaining it:
  - `public/` is excluded: static SVG assets, not source. Biome 2.5 parses `.svg` files and fires `noSvgWithoutTitle` on them, but their accessible text lives on the `` that references them.
  - `noImportantStyles` is off for `app/global.css` only (file-scoped `overrides`), because that file exists to override the fumadocs theme's own stylesheet.
  - `css.parser.tailwindDirectives: true` so the `@theme {}` block parses instead of erroring.
- **Suppress, with a reason on each line**, `noDangerouslySetInnerHtml` on the two static analytics bootstraps (Plausible, reb2b) and `noImgElement` on the Scarf tracking pixel. Both snippets are constant strings with no user input. Moving them to `next/script` would silence the rule but change when the analytics scripts execute, which does not belong in a lint-adoption PR.
- **Fix the real findings**: `type="button"` on the copy-markdown button (default is `submit`), content keys instead of array indexes in the two diagram components, template literals in `convert-sdk-docs.mjs` with a small `writeJson` helper for the two `meta.json` writes.
- **CI**: add a `Lint docs app` step to `docs.yml`, placed right after `pnpm install` and *before* `setup-uv` / `uv sync`, so a lint failure fails in seconds rather than after the Python toolchain install. Add a `just docs-lint` recipe and note in `docs/CLAUDE.md` that root `just check` does not cover it.

## Reviewer Notes

The diff is large but almost all of it is the formatter. The parts worth actual attention:

1. **`docs/biome.jsonc`** is the policy file. Every exception is a deliberate decision with a comment; if you disagree with any of them, that is the place to push back. `includes` was trimmed to `["**", "!public"]` because `vcs.useIgnoreFile: true` already makes biome honour `docs/.gitignore` (`node_modules`, `.next`, `.source`, `build`, `out`, generated `content/docs/{cli,reference}`).
2. **`docs/app/layout.tsx`** carries the three per-line suppressions. If the analytics snippets are wrong, they were wrong before; this PR only annotates them.
3. **`docs/components/diagrams.tsx`** switches `key={i}` → `key={b}` / `key={i.label}`. Every `bullets` and `items` value in that file is a static literal array with no duplicate strings, so this is safe today. If a future diagram repeats a bullet, React will warn about a duplicate key and drop the second item; that is now an ordinary content-authoring constraint rather than a hidden one.
4. **`.github/workflows/docs.yml`** step reorder: the pnpm block moved above the uv block. No later step depends on the relative order of the two toolchains; the generate steps still run after both are installed.
5. **`docs/scripts/convert-sdk-docs.mjs`**: the `useTemplate` auto-fix produced a template literal wrapping a multi-line `JSON.stringify(...)` that read worse than the original, so both `meta.json` writes now go through `writeJson(path, value)`. Output is byte-identical (checked by running `just generate-docs`).

What breaks if this is wrong: a mis-scoped `includes` would lint generated MDX or shipped SVGs and go red on the next docs-generation change; a wrong `overrides` path would let `!important` through in any future stylesheet; a bad step order in `docs.yml` would fail the docs build for every PR touching `docs/`.

### Reproduction

```bash
# On develop: the script exists but fails
git checkout develop && cd docs && pnpm install --frozen-lockfile && pnpm run lint
# → Found 41 errors.

# On this branch: clean
git checkout docs-biome-lint-ci && cd docs && pnpm install --frozen-lockfile && pnpm run lint
# → Checked 33 files. No fixes applied.

# Same thing via the new recipe, from the repo root
just docs-lint

# Prove global.css is semantically unchanged (quotes, hex case, whitespace normalised)
git show develop:docs/app/global.css | tr "'" '"' | tr 'A-F' 'a-f' | tr -s ' \n' > /tmp/old.css
tr "'" '"' < docs/app/global.css | tr 'A-F' 'a-f' | tr -s ' \n' > /tmp/new.css
diff /tmp/old.css /tmp/new.css   # only line-wrap differences

# The site still builds and the SDK reference still generates
just generate-docs && cd docs && pnpm run types:check && pnpm run build
```

In CI, look for the new `Lint docs app` step in the `SDK Reference Docs` job on this PR; it should run before `Generate CLI reference pages` and pass.

Local checks run: `just check`, `just test` (3730 passed), `just zizmor`, `just docs-lint`, `just generate-docs`, docs `types:check` and `build`.

## Follow-ups noticed, not addressed here

- `just docs-validate` calls `pnpm run validate:export`, which does not exist in `docs/package.json` (stale recipe).
- Root `biome.json` (2.3.11, `rules.recommended`) and `docs/biome.jsonc` (2.5.10, `rules.preset`) are independent installs; rule behaviour can drift between them. Unifying means pulling `docs/` into the root pnpm workspace, which is a separate decision.


AlexejPenner (approved):
🦭

PR #889, opened 2026-08-25
Optionally scope evaluators to an agent

### LINKED ISSUE #669: Link Evaluators to Agent
Evaluators can be quite specificly designed for one specific agent or even one specific batch of sessions/cohort - it may make sense to formalize this relationship

schustmi:
Adds an optional `agent_id` to evaluator creation, null for a workspace-global evaluator. A scoped evaluator can only be used by experiments, replays, and evaluation batches for that agent, and evaluator listing supports filtering on `agent_id`. Evaluation batches now require every input session to belong to the same agent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Added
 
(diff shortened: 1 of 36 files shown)


AlexejPenner:
before we merge this - what effect does this have on the UI? breaking change or not?

AlexejPenner (approved):
🦭

PR #895, opened 2026-08-26
Harden core release automation

AlexejPenner:
## Summary

- add deterministic core release version proposal and candidate validation based on merged PR labels
- reject development manifests as release baselines
- add a deterministic post-release reset command for the core manifest, both lockfiles, generated OpenAPI version, and changelog
- have every successful stable core tag create a draft development-reset PR against `develop`
- require release tags to be pushed individually and verified one workflow run at a time
- update both release-skill mirrors with the enforced version and post-release workflow

## Why

Release preparation previously left the core version decision and post-release reset in prose. That allowed the development manifest version to outweigh the `Breaking Change` labels, and it allowed the required `+dev` reset to be forgotten after publication.

The release inventory script now owns both transformations. For example, this validates the breaking release proposal:

```bash
uv run --no-project --with packaging==26.2 \
  python scripts/release_units.py propose-core-version \
  --latest-version 0.22.2 \
  --label "Breaking Change" \
  --candidate 0.23.0 \
  --format json
```

After a successful stable core tag, `release.yml` uses `RELEASE_GIT_TOKEN` to create a draft `Start Kitaru <version> development` PR. The deterministic reset sets `<version>+dev`, updates both lockfiles and the generated OpenAPI version, and restores the top `Unreleased` section. The PR remains draft until `main` contains the clean release commit.

## Validation

- `uv run pytest -q tests/scripts/test_release_units.py tests/cli/test_repo_skills.py` (87 passed)
- `just check`
- exercised the `0.23.0` reset in an isolated repository copy
- `uv lock --check` and `uv lock --project plugins --check` after the reset
- skill validation for both `.agents` and `.claude` copies
- byte-for-byte skill mirror comparison
- `git diff --check`

## Reviewer Notes

Review `scripts/release_units.py` and its focused tests first: the version rule and reset transformation fail before writing when repository state is inconsistent. Then review the final `create-development-reset-pr` job in `release.yml`: it runs only for stable tag pushes after the release and maintenance-branch jobs, uses the release identity so normal PR CI triggers, handles workflow retries, and creates a draft PR changing only `pyproject.toml`, `uv.lock`, `plugins/uv.lock`, `openapi/openapi.json`, and `CHANGELOG.md`. Finally verify the two release-skill copies are identical.


strickvl (approved):
I do think the whole 'figure out what release number' is a bit overdone. I guess there are lots of situations where we might want to override things, but I assume you can do that by just typing in the agent.

Also I assume it will confirm with the user what version number(s) to use? If not, we should add that before merging.

AlexejPenner:
> I do think the whole 'figure out what release number' is a bit overdone. I guess there are lots of situations where we might want to override things, but I assume you can do that by just typing in the agent.
> 
> Also I assume it will confirm with the user what version number(s) to use? If not, we should add that before merging.

Yeah you can always instruct the skill to do version 99.99.9 if you like

PR #896, opened 2026-08-26
Split session responses into slim and detail models

schustmi:
Session list, create, update, and node ingest responses no longer carry payloads. Fetch them via the new detail-shaped GET /api/v1/sessions/{session_id}.


AlexejPenner (approved):
🦭

AlexejPenner:
FE changes required? if so pls add the label

schustmi:
> FE changes required? if so pls add the label

nope, not required

PR #903, opened 2026-08-28
Remove duplicate `spellcheck.yml` workflow

strickvl:
## What

Deletes `.github/workflows/spellcheck.yml`. No other changes.

## Why

`ci.yml` already has a `typos` job that is a required status check on `develop`. `spellcheck.yml` ran the same scan again on every push and PR, but was not required, so it added CI minutes without adding any protection.

Verified the two jobs are identical, not just similar:

| | `ci.yml` → `typos` | `spellcheck.yml` → `spellcheck` |
|---|---|---|
| Action + SHA | `crate-ci/typos@8a48f81…` (v1.49.0) | same |
| Triggers | push to `develop`, PRs, skips drafts | same (+ `workflow_dispatch`, `workflow_call`) |
| Inputs | defaults | `files: .`, `config: ./.typos.toml` |
| Required check | yes | no |

The explicit inputs in `spellcheck.yml` equal the action's defaults: `files` defaults to `.`, and the `typos` binary auto-discovers `.typos.toml` in the repo root. So both jobs scanned the same files with the same config.

Nothing referenced `spellcheck.yml` via `workflow_call`, and the `develop` ruleset only lists `typos` (not `spellcheck`), so no ruleset edit is needed. `just typos` remains the local equivalent.

## Reviewer Notes

The only risk would be if the two jobs were *not* actually equivalent, in which case deleting one would drop coverage. The table above is the argument that they are. The one thing worth eyeballing: confirm `.typos.toml` is at the repo root (it is), since that is what makes the implicit config discovery in `ci.yml` match the explicit `config:` in the deleted file.

### Reproduction

On any recent PR, open the Checks tab: both `typos` and `spellcheck` appear, both green, both running `crate-ci/typos`. After this merges only `typos` appears, and it is still the required one.

Local checks run: `just actions-lint`, `just zizmor` (no findings).


AlexejPenner (approved):
🦭

PR #911, opened 2026-08-31
Support OpenAI Agents SDK 0.20-0.22

strickvl:
## What changed?

- `kitaru-openai-agents` now accepts `openai-agents>=0.19.3,<0.23` instead of `<0.20`.
- `plugins/uv.lock` resolves `openai-agents` 0.22.0 and, as required by the 0.21 provider update, `openai` 3.5.0 (previously 2.53.0).
- `test_supports_locked_openai_agents_minor_line` guards the new upper bound.

No adapter source change was needed.

## Why?

Weekly upstream compatibility check. Latest released `openai-agents` is 0.22.0, three minor lines ahead of the pinned bound, so anyone installing the adapter alongside a current OpenAI Agents SDK hit a resolution failure.

The adapter's whole surface passes unchanged against 0.22.0 in a fresh env: 97 of 98 tests passed, and the only failure was the bound assertion itself (`assert Version('0.22.0') < Version('0.20')`). Notably `test_matches_complete_public_runner_signatures` and `test_locks_supported_hook_methods` — which mirror `Runner.run`/`run_sync` signatures and the `RunHooks`/`AgentHooks` method sets — still match, so the wrapped API did not move.

Upstream release notes for the three skipped minors contain no change to the surfaces this adapter wraps:

- 0.20.0 / 0.21.0: new `agents.testing` utilities plus the OpenAI provider move to `openai>=3.0.0,<4` (this is what forces the `openai` bump here); no breaking SDK behavior change per the release notes.
- 0.21.1: model call timeouts, sandbox/realtime/session fixes.
- 0.22.0: guardrail redaction of terminal tool output, `ModelBehaviorError` for terminal `failed`/`incomplete` non-streaming responses, `RunState` usage isolation. The one tightened contract is `OpenAIProvider` rejecting `organization`/`project` alongside an explicit `openai_client`, which the adapter never sets.

## Release context

Requires a `kitaru-openai-agents` release for users to pick up the widened bound. `openai` 3.5.0 in `plugins/uv.lock` is dev-only; the root workspace lockfile is untouched.

A follow-up PR widens `kitaru-pydantic-ai` to the current `pydantic-ai-slim` line. That lock update is only resolvable once this one lands, because `pydantic-ai-slim>=2.32` needs `openai>=3` and `openai-agents<0.20` caps `openai` at 2.x.

## Reviewer Notes

The risk here is not the diff, it is the claim that three upstream minors changed nothing this adapter depends on. Two tests carry that weight, so read them rather than the bound: `plugins/tests/adapters/openai_agents/test_runner.py::test_matches_complete_public_runner_signatures` compares `KitaruRunner.run`/`run_sync` parameter names against the upstream `Runner` methods, and `test_locks_supported_hook_methods` locks the hook method sets. If upstream had renamed or added a runner parameter or hook, those fail — they do not.

The second thing to weigh is `openai` 2.53.0 → 3.5.0 in the plugin workspace lock. That is a major bump pulled in transitively by `openai-agents` 0.21+, and it applies to every plugin package resolved in that workspace, not just this adapter. The full plugin suite (480 tests, including the importers and evaluator) passes on it, and `just plugin-artifact-smoke` builds and installs the wheels cleanly.

## Reproduction

Install the adapter against the latest upstream release in a throwaway env and run its suite:

```bash
cd ~/repos/kitaru
uv sync --project plugins --frozen --all-packages
uv run --project plugins python -c "from importlib.metadata import version; print(version('openai-agents'), version('openai'))"
# 0.22.0 3.5.0
uv run --project plugins pytest -q -c plugins/pyproject.toml plugins/tests/adapters/openai_agents
```

Expect 98 passed, including the recording and replay suites (`test_recording.py`, `test_replay.py`) that cover tool-call capture and history replay. On `develop` the same commands resolve `openai-agents` 0.19.4, and forcing 0.22.0 there fails resolution with `kitaru-openai-agents depends on openai-agents>=0.19.3,<0.20`.

## Local checks run

- `uv run --project plugins pytest -q -c plugins/pyproject.toml plugins/tests tests/server/test_default_plugins.py` — 480 passed
- `uv run --project plugins ruff format --check` / `ruff check` / `ty check --project plugins` — clean
- `just plugin-artifact-smoke` — passed

Link to Devin session: https://app.devin.ai/sessions/3c67e268b3d4404c90dd04d3606dd39d
Open in Devin Desktop: https://app.devin.ai/desktop/session/3c67e268b3d4404c90dd04d3606dd39d?variant=devin
Requested by: @strickvl


AlexejPenner (approved):
🦭

PR #917, opened 2026-08-31
Importer-backed adapters

### LINKED ISSUE #869: Implement a generic adapter
Implement a generic adapter that runs an importer inline, so harnesses without a dedicated adapter can record sessions: the harness runs natively, and its trace output is converted to Kitaru sessions through the importer at runtime instead of requiring a bespoke per-framework adapter.

schustmi:
This PR adds importer-backed adapters: wrap an agent entrypoint that is already instrumented with an observability provider, and the adapter imports the provider trace of each run into Kitaru as a session. Adapters ship inside the provider importer packages for Langfuse, Braintrust, LangSmith, Logfire, and Phoenix behind an `adapter` extra, on top of the `ImporterBackedAdapter` base class in core. Run specs declare `runtime_capabilities`, and creating a replay or starting an experiment run rejects overrides and tool policies the agent version does not declare.

**Example:**

```python
# agent.py
from kitaru_langfuse_importer.adapter import LangfuseAdapter

def run_agent(question: str) -> str:
    ...  # the Langfuse-instrumented agent

adapter = LangfuseAdapter()
result = adapter.run(run_agent, question="What is an AI agent?")
```

Register the entrypoint as an agent version to target it with session runs and replays:

```bash
kitaru agent version register langfuse-agent --spec version.yaml
```

```yaml
# version.yaml
run_spec:
  command: python agent.py
  runtime_capabilities:
    overrides: false
    tool_policies: false
```


AlexejPenner (approved):
🦭

strickvl (approved):

PR #922, opened 2026-08-31
Evaluation params and statistics

schustmi:
This PR turns evaluations into append-only measurements that record how they were produced and pins experiment run statistics to the evaluations that scored them.

- Create replays and experiment runs with `baseline_evaluation_mode` (`none`, `if_missing`, `force`), defaulting to `if_missing`. `evaluate_baselines` survives on the REST API only, deprecated.
- Evaluator-born rows store the params they ran with and keep their evaluator version id after the evaluator is deleted. Manual evaluations are unique per session and name, duplicates return HTTP 409.
- A replay-evaluation link records which measurements a replay's comparison uses, written when a task produces rows or when `if_missing` adopts the latest evaluation matching session, evaluator version, and params.
- Run aggregates read only linked evaluations and group per evaluator version, so a later evaluation never changes an existing run's numbers.
- Evaluation results take optional `min_score`, `max_score`, and `target_score` on float results for scale display.
- `DEPRECATIONS.md` tracks deprecated surfaces.


DEPRECATIONS.md
strickvl, line 1: maybe leave a oneliner in AGENTS.md / CLAUDE.md that this file exists, so that they know to update it?

strickvl (approved):
skills will also need updating for sure once this lands.

schustmi:
> skills will also need updating for sure once this lands.

Yep, created a ticket for it already

PR #925, opened 2026-08-31
Fix redaction and capture invariants

### LINKED ISSUE #906: Redaction and capture invariants: silent key collapse, plural secret keys, unbounded recursion
While setting up property-based tests, two functions turned out to violate invariants they are relied on for. Both share a shape: a dict is rebuilt with `str(key)`, so distinct keys that stringify identically collapse silently.

### 1. `redact_data` (`src/kitaru/cli/redaction.py`, duplicated in `src/kitaru/mcp/redaction.py`)

Invariants it should hold: (a) no secret value survives into output; (b) `len(redact_data(d)) == len(d)` for dicts; (c) never raises. Verified results:

```python
from kitaru.cli.redaction import redact_data

redact_data({1: "a", "1": "b"})
# -> {'1': 'b'}                       (b) violated: silent key loss

redact_data({"apiKey": "s1", "tokens": "s2", "x-api-key": "s3", "API_KEY": "s4", "secrets": {"a": "s5"}})
# -> {'apiKey': '***', 'tokens': 's2', 'x-api-key': '***', 'API_KEY': '***', 'secrets': {'a': 's5'}}
# (a) partially violated: plural forms `tokens` / `secrets` pass through

d = {}; cur = d
for _ in range(3000):
    cur["k"] = {}; cur = cur["k"]
redact_data(d)
# -> RecursionError                    (c) violated
```

The recursion case matters more than it looks: `redact_data` runs on the CLI result-emission path (`src/kitaru/cli/output.py` `_write_json`, `_detail_value`, `_display_value`). A `RecursionError` there is caught by the `except BaseException` in `cli/app.py`, which calls `emit_error`, which calls `redact_data` again on the error details. If that second call also recurses, it escapes `main()` as a bare traceback.

The plural-key case is partly a decision rather than a bug: `tokens` is also a legitimate token-usage key. The regex's notion of "secret key" is undertested; the property tests pin the current behavior as a strict xfail so the decision has to be made explicitly.

Both `_SECRET_KEY` and `_SECRET_VALUE` regexes are linear (no ReDoS).

### 2. LangGraph `capture_value` reports `replayable=True` on data it dropped

`plugins/packages/langgraph/src/kitaru_langgraph/capture.py` `capture_value` (the `str(key)` coercion is in `_convert`) and `codec.py` `_encode_nested` share the same shape.

```python
from kitaru_langgraph.capture import capture_value, CapturePolicy
r = capture_value({1: "a", "1": "b"}, CapturePolicy())
print(r.value, r.lossy, r.reasons, r.replayable)
# -> {'1': 'b'} False () True
```

One key wins, the other is dropped, nothing is appended to `reasons`, `lossy` stays `False`, and `replayable` claims the stored copy is exact. Replay depends on that flag being honest. The property to hold: for any mapping `d`, either `len(capture_value(d, p).value) == len(d)`, or `lossy is True` with a non-empty `reasons` and `replayable is False`. Same for `{True: 'a', 1: 'b'}` and `{(1,): 'x', '(1,)': 'y'}`.

Related, unconfirmed: the `encode_tool_outcome` / `decode_tool_outcome` round-trip in `codec.py` (tuples become lists; `status` silently defaults to `"success"`), and `decode_tool_outcome` only converts `TypeError` / `ValueError` into `ToolPolicyError`, so anything else raised on a hostile stored envelope escapes.

### Tests

The property-based testing PR adds `tests/cli/test_redaction_properties.py` and `plugins/tests/adapters/langgraph/test_capture_properties.py` with the invariants above as properties and the confirmed cases as `xfail(strict=True)` regressions pointing here.

strickvl:
## Summary

CLI and MCP output now masks the same credential fields, preserves distinct mapping entries, and stays serializable for deeply nested or cyclic values. LangGraph recording now rejects captured outcomes whose mapping keys changed, preserves tuples, and fails closed on malformed stored results instead of executing the live tool.

The redaction exception for token usage is deliberately narrow: only numeric or null values in the current `TokenUsage` shape remain visible. Mixed or credential-shaped token fields are masked.

## Reviewer Notes

Please focus on the mixed-key suffix policy in `kitaru.redaction`, the token-usage exception, and backward compatibility of the existing `kitaru.langgraph.tool_result.v1` envelope. New tuple tags fail closed on older adapter versions; existing list values remain lists.

### Reproduction

1. Run `uv run pytest -q tests/cli/test_redaction_properties.py tests/cli/test_output.py tests/mcp/test_handlers_protocol.py tests/mcp/test_settings_errors.py`.
2. Run `uv run --project plugins pytest -q -c plugins/pyproject.toml plugins/tests/adapters/langgraph`.
3. Confirm the #906 tests are ordinary passing tests and no longer carry strict xfail markers.

Local validation: `just check`; full core suite (4,167 passed, 16 skipped, 14 unrelated xfailed); full plugin suite (578 passed, 14 unrelated xfailed); relevant CI/nightly Hypothesis profiles; JavaScript package tests (102 passed); Python TypeScript integration tests (15 passed).

Code review: harness-native fallback. The configured peer review was blocked by the host approval boundary, so Codex correctness, adversarial, testing, and project-standards reviewers covered the diff locally. All four findings were applied and their affected tests rerun.

- Search for `The MCP result could not be serialized` and `Stored tool result payload is malformed` during the first seven days after release.
- Healthy signal: CLI/MCP result emission remains successful and valid LangGraph history hits continue to replay without new policy errors.
- Failure signal: a sustained increase in serialization internal errors or rejected previously valid LangGraph history. If seen, revert this PR from the release branch and inspect the rejected payload shape before re-releasing.
- Owner: Kitaru maintainers.

Closes #906

---

[Compound Engineering](https://github.com/EveryInc/compound-engineering-plugin)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ### Fixed
 
+- Unified CLI and MCP credential masking, including plural credential fields, while preserving numeric token usage. Output now preserves colliding mapping keys and safely bounds deep or cyclic data.
+- LangGraph capture now marks mapping-key coercion and collisions as non-replayable. Tool-result encoding preserves nested tuples, and replay rejects missing message status and malformed stored outcomes without running the live tool.
 - Fixed `kitaru evaluator version register` and `kitaru importer version register`, which always failed with a `TypeError` after uploading the source. `POST /api/v1/evaluators/{evaluator_id}/versions` and `POST /api/v1/importers/{importer_id}/versions` now support idempotency keys, and the SDK `create_version` methods take an `idempotency_key` argument, matching agent and cohort version creation. The MCP evaluator management tool's `create_version` operation takes an optional `idempotency_key` field.
 
 ## [0.23.0]
--- a/docs/book/adapters/langgraph.md
+++ b/docs/book/adapters/langgraph.md
@@ -112,6 +112,10 @@ Misses follow the replay policy without silent fallback:
 
 Malformed, unsupported, or lossy recorded results fail closed with `ToolPolicyError`; they do not become misses or permit passthrough. A matched recorded failure also raises `ToolPolicyError` with the stored error text and aborts the graph unless application middleware catches it. Kitaru does not recreate the original exception class or LangGraph recovery behavior. The adapter rejects the `llm` tool policy. It does not substitute stored model responses.
 
+Newly recorded supported tool results preserve nested tuples and lists as distinct types, including tuple pairs in `Command.update` and the default empty tuple in `Command.goto`. Stored `ToolMessage` values must have an explicit `success` or `error` status; a missing or invalid status is rejected rather than treated as success.
+
+The tool-result envelope remains `kitaru.langgraph.tool_result.v1`. Valid older envelopes remain readable, and their stored lists remain lists. Older adapter versions reject the new nested tuple tag, so use an updated adapter to replay newly recorded tuple-bearing results. These changes cannot recover keys or tuple types already lost in historical records.
+
 ## Interrupts and unsupported invocation modes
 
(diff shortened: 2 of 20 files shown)


AlexejPenner (approved):
🦭

PR #948, opened 2026-09-01
Document the string-detail 422 body in the OpenAPI spec

### LINKED ISSUE #930: OpenAPI declares only the array-shaped 422 body, but hand-raised validation errors return a string `detail`
Found by the Schemathesis API fuzzing spike (#923). Verified by hand against the real app.

## What happens

Kitaru returns two different shapes of 422 body, and `openapi/openapi.json` documents only one of them.

The handler at `src/kitaru/server/api/app.py:147-150` serializes every domain validation error as a plain string:

```json
{"detail": "Invalid cursor"}
```

FastAPI's OpenAPI generation knows nothing about that handler, so it documents only its own auto-generated 422 body, `HTTPValidationError`, whose `detail` is an **array** of `ValidationError` objects. That `$ref` is the sole 422 schema in the entire spec — verified across all 141 operations.

Both shapes really do occur on the same route. FastAPI's own request-validation 422 still returns the documented array shape:

```
GET /api/v1/accounts?limit=notanint
→ 422 {"detail": [{"type": "extra_forbidden", ...}]}    # documented array shape

GET /api/v1/accounts?cursor=
→ 422 {"detail": "Invalid cursor"}                      # undocumented string shape
```

So the true contract is a union of the two, and the spec declares only half of it.

## Scope

This single signature caused **43 of the 47** failures in the fuzzing spike's positive pass. It fires on essentially every resource: accounts, agents, agent versions, annotations, api-keys, cohorts, cohort versions, devices, evaluations, evaluators, evaluator versions, experiments, experiment-runs, experiment-run jobs, importers, importer versions, investigations, jobs, job tasks, replays, secrets, service-accounts, sessions, session nodes, tags, tasks, ui/sample-data, ui/sessions, users, workers.

Other verified instances of the string shape:

```
GET  /api/v1/jobs?sort=j:asc                          → {"detail": "Invalid sort 'j:asc'"}
POST /api/v1/agents  {"name": "", "description": null} → {"detail": "Name must not be empty"}
```

## Why the fix belongs in the spec, not the handler

The string shape is deliberate and is the contract Kitaru's own SDK depends on — `src/kitaru/client/exceptions.py:102` parses `detail` as a string. Changing the handler to emit the array shape would break the shipped SDK. The server is right; the document is wrong.

## Impact

No crash and no data exposure, but any third-party client generated from `openapi/openapi.json` will fail to parse every domain validation error Kitaru actually returns — which is the large majority of them.

## Suggested fix

Add a small `ErrorResponse` model (`{"detail": str}`) to `api_models`, then rewrite every operation's 422 response schema to `anyOf: [HTTPValidationError, ErrorResponse]` in one central place. Two options:

- a `custom_openapi` override in `create_app`, or
- a post-processing step in `scripts/generate_openapi.py`

The second is simpler, since `openapi.json` is already generated there and staleness-gated by `scripts/check_openapi.sh`.

While in there, the same treatment applies to the other hand-raised string-detail statuses — 403, 404, 409, 413, 426, 503 — which the spec mostly does not declare at all.

Regenerate `openapi/openapi.json` and run `pnpm run generate` for the TypeScript types, since the union changes generated client types.

## Related coverage gap worth folding in or splitting out

The spike hit a second, larger documentation hole that blocks a whole class of fuzzing checks. Across the entire spec, the complete set of declared status codes is `{200, 201, 204, 400, 422}` — every operation declares its success code plus 422 and nothing else. Every legitimate 404, 401, 403, 409, and 503 the server returns is undocumented.

The practical consequence: Schemathesis's `status_code_conformance` check is unusable against this spec, because it flags a correct 404 on `DELETE /agents/{id}` as a violation. That cost the spike its stateful testing run until the check was removed. Fixing the 422 union without also declaring the other statuses leaves that check off the table.

schustmi:
Fixes #930

### DIFF
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -3177,19 +3177,6 @@
         "title": "FilterOp",
         "type": "string"
       },
-      "HTTPValidationError": {
-        "properties": {
-          "detail": {
-            "items": {
-              "$ref": "#/components/schemas/ValidationError"
-            },
-            "title": "Detail",
-            "type": "array"
-          }
-        },
-        "title": "HTTPValidationError",
-        "type": "object"
-      },
       "HistoryConfig": {
         "additionalProperties": false,
         "description": "History tool config.",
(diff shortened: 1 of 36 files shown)


AlexejPenner (approved):
🦭

strickvl (approved):

PR #949, opened 2026-09-01
Return OAuth error codes on login validation failures

### LINKED ISSUE #929: `POST /api/v1/login` validation errors omit the required `error` field from their declared 400 body
Found by the Schemathesis API fuzzing spike (#923). Verified by hand against the real app.

## What happens

`POST /api/v1/login` declares its 400 response as `TokenErrorResponse`, which requires an `error` field carrying an OAuth 2.0 error code. The device-grant failures produce that shape correctly. The request-validation failures do not — they raise a bare `HTTPException(400, detail=...)`, which FastAPI renders as `{"detail": "..."}` with no `error` key.

```
POST /api/v1/login
Content-Type: application/x-www-form-urlencoded

device_id=e3e70682-c209-1cac-a29f-6fbed82c07cd&username=
```

→ `400 {"detail": "Invalid request: username is required."}` — missing the required `error` property.

Siblings reproduce identically:

- device grant without `device_code` → `{"detail": "Invalid request: device_id and device_code are required."}`
- `grant_type=bogus` → `{"detail": "Unsupported grant type: bogus"}`

Control, showing the conforming shape exists: device grant with a wrong code → `400 {"error": "invalid_grant", "detail": "Invalid device code"}`.

## Why this is a server defect rather than a spec gap

Two independent signals say the spec documents the intended behavior and the handler falls short of it:

1. `TokenErrorCode` already defines `INVALID_REQUEST` and `UNSUPPORTED_GRANT_TYPE` (`src/kitaru/api_models/v1/auth.py:44-45`) and the server never uses either. They were clearly meant for exactly these branches.
2. RFC 6749 §5.2 requires an `error` field on a token endpoint's 400 response. Third-party OAuth clients branch on it.

`/api/v1/login` is the only endpoint in the whole spec that declares a 400 at all, so this is the one place the OAuth error contract is written down, and it is not honored.

## Affected branches

All in `src/kitaru/server/adapters/rest/routers/auth.py`, inside the `LoginRequestForm` dependency:

- `:123` — missing username
- `:137` — missing `device_id`/`device_code`
- `:156` / `:173` — unknown or unsupported `grant_type`
- `:166` — missing grant type

The route decorator declaring `TokenErrorResponse` for 400 is at `auth.py:234`. The conforming path is `auth.py:334-361` raising `TokenGrantError`, converted by the handler at `src/kitaru/server/api/app.py:219`.

## Impact

Low. Correct status code, informative message, and Kitaru's own SDK is unaffected — `src/kitaru/client/exceptions.py:104-112` already prefers the `error` key when present and falls back to `detail`. The cost is spec conformance and third-party OAuth clients that branch on `error`.

## Suggested fix

Make `LoginRequestForm` raise `TokenGrantError` instead of bare `HTTPException`:

- missing username → `TokenGrantError(TokenErrorCode.INVALID_REQUEST, ...)`
- missing device fields → `INVALID_REQUEST`
- `_unsupported_grant_type` → `UNSUPPORTED_GRANT_TYPE`
- missing grant type → `INVALID_REQUEST`

No handler change is needed: the `TokenGrantError` handler already registered in `create_app` fires for exceptions raised inside FastAPI dependencies too.

Do **not** instead widen the OpenAPI 400 schema to a union — that would document RFC 6749 non-compliance rather than fix it.

## Regression tests

Assert the `error` value for each validation branch: empty username → `invalid_request`, `grant_type=bogus` → `unsupported_grant_type`, device grant without `device_code` → `invalid_request`.

## One judgment call for the PR

`POST /api/v1/device_authorization` has a similar undeclared bare 400 ("This server does not authenticate requests."). Worth either folding into this change or splitting into its own ticket.

schustmi:
Fixes #929

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 - A NUL byte or another C0 control character in a user-supplied string now returns HTTP 422 instead of HTTP 500. Request string fields reject the characters at validation, and a database error caused by a value the database cannot store maps to HTTP 422 as a backstop.
 - Getting or updating an evaluator or importer version with a version number outside PostgreSQL's 32-bit integer range now returns HTTP 422 instead of HTTP 500.
 - Reusing an `Idempotency-Key` registered on another route against `POST /api/v1/api-keys` or `POST /api/v1/api-keys/{api_key_id}/rotate` now returns HTTP 422 instead of HTTP 500. The stored response is only decrypted after the request fingerprint matches, and a stored response that cannot be decrypted returns HTTP 409.
+- `POST /api/v1/login` validation failures now carry the OAuth 2.0 `error` code their declared 400 response requires, `invalid_request` for a missing field and `unsupported_grant_type` for an unknown or rejected grant type. `POST /api/v1/device_authorization` declares its 400 response and returns the same shape when the server does not authenticate requests.
 
 ## [0.24.0]
 
--- a/openapi/openapi.json
+++ b/openapi/openapi.json
@@ -11646,7 +11646,7 @@
     },
     "/api/v1/device_authorization": {
       "post": {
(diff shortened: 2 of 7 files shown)


AlexejPenner (approved):
🦭

PR #952, opened 2026-09-02
Add one-line installer served at kitaru.ai/install

htahir1:
## Summary

Adds a one-line installer so getting Kitaru onto a machine is a single command instead of a four-step README:

```bash
curl -fsSL https://kitaru.ai/install | bash
```

The script installs the CLI and MCP server in an isolated `uv` tool environment (installing `uv` first if needed, with a uv-managed Python when the machine has none), installs the coding-agent skills for every agent it finds, registers the Kitaru MCP server with Claude Code and Codex, and starts the local server with `kitaru login --local` when Docker is running and a terminal is attached. Nothing needs sudo; everything lands under `$HOME`; re-running upgrades.

The kitaru.ai redirect worker gains one content route, `/install` (and `/install.sh`), that proxies `install.sh` from `main` so the one-liner resolves on our domain without a redirect hop.

## Reviewer Notes

**The shape.** Kitaru is a Python package, not a binary, so this is deliberately not a manifest-and-checksum binary installer. `uv tool install "kitaru[cli,mcp,worker]"` is the whole install; the script's job is the plumbing around it. That keeps the installer boring and means a release is just a PyPI publish.

**Where the risk lives:**

- `install.sh` is wrapped in `main "$@"` and every non-interactive child runs with `</dev/null`. Both matter under `curl | bash`: without them a child that reads stdin swallows the rest of the script mid-stream. I hit exactly this with `npx skills` before adding the guard.
- The skills CLI (`npx skills add`) exits 0 even when it installs nothing, so the script checks for `kitaru-investigation/SKILL.md` on disk afterwards and falls back to a tarball copy from GitHub (no git, no Node needed) into `~/.agents/skills` plus `~/.claude` / `~/.codex` when present. Those are the user-scope paths `kitaru.cli.skill_discovery` searches, so `kitaru doctor` sees them.
- MCP registration uses the absolute path to `kitaru-mcp` in uv's tool bin dir rather than relying on PATH, since GUI-launched agents often don't have `~/.local/bin`. Existing `kitaru` entries are replaced, not duplicated.
- Login only runs with a TTY (reads from `/dev/tty` because stdin is the pipe). No Docker, or no TTY, prints the one command to run next instead of failing.
- `docs/worker/redirect.mjs`: `/install` is the only path on kitaru.ai that returns content. It fetches raw GitHub `main` with a 5-minute edge cache and returns 502 with a docs pointer if GitHub is down. Everything else still redirects.

**What would break if wrong:** a user runs the one-liner and ends up with a half-configured machine and no clear next step. The script's failure modes all end in a printed command, never a silent partial state.

**Deploy note:** the worker deploys from `main` via `docs.yml`, and the route fetches `install.sh` from `main`. Until the next develop-to-main release the URL 404s, so the README link is excluded from lychee for now (comment in `lychee.toml` says why; remove once live).

### Reproduction

Isolated end-to-end run, exactly as a new user would hit it, with a throwaway home:

```bash
export H=$(mktemp -d); mkdir -p "$H/.claude"
cat install.sh | env -i HOME="$H" PATH="/usr/bin:/bin:$HOME/.local/bin:$(dirname "$(command -v npx)")" bash -s
ls "$H/.local/bin"            # kitaru, kitaru-mcp
ls "$H/.agents/skills"        # six kitaru-* skills
python3 -c "import json;print(json.load(open('$H/.claude.json'))['mcpServers']['kitaru'])"
HOME="$H" "$H/.local/bin/kitaru" doctor
```

Without Node on PATH the same run takes the tarball path and prints "Skills copied into ~/.agents/skills". With Docker running and a real terminal, the last step starts the local server and logs in.

Docker matrix, each image piping the script through `bash -s -- --no-login` like curl would, then checking `kitaru --version`, `kitaru-mcp`, six skills on disk, `kitaru doctor`, and that PATH was persisted in an rc file:

| Image | Arch | Result |
|---|---|---|
| ubuntu:24.04 | amd64 | pass |
| debian:bookworm-slim | amd64 | pass |
| fedora:41 | amd64, arm64 | pass |
| archlinux | amd64 | pass |
| alpine:3.21 (musl) | amd64, arm64 | pass |
| node:22-bookworm-slim (exercises `npx skills`) | amd64 | pass |
| python:3.11-slim (system Python present) | amd64 | pass |

Two installer bugs came out of the matrix and are fixed in this PR: the uv bootstrap was piped into a helper that closes stdin (failed on every image without uv), and PATH was not persisted on shells uv does not recognise (Alpine). Real-OS runs on GitHub-hosted runners via the new `installer.yml` workflow (runs on any PR touching `install.sh`): ubuntu-latest, ubuntu-24.04-arm, macos-latest (Apple Silicon), macos-15-intel, and windows-latest under Git Bash, all green: https://github.com/zenml-io/kitaru/actions/workflows/installer.yml. Each job pipes the script through `bash -s`, verifies CLI, MCP binary, skills, `kitaru doctor`, and PATH persistence, then re-runs the script to prove a second run is an upgrade. Windows/Git Bash was the surprise: uv, the CLI, the MCP server, and the skills all install; only the local server needs Docker Desktop, so the script now proceeds there with a warning instead of refusing.

Docs: the installation page was restructured around one install path (curl line, what it does, verify, local server, team server), with project-environment installs, TypeScript, and skills-only under "Other ways to install".

Worker: `node --test docs/worker/redirect.test.mjs` (17 pass, including `/install`, `/install.sh`, `/install/`, and the 502 path).

Local checks run: the Docker matrix above, worker tests, `typos` on the changed files, `bash -n`. `just` is not on this machine so the full `just check` was not run; the changes are shell, JS, and Markdown only.


htahir1:
Follow-up filed: #953 (`kitaru setup` command to own the skills + MCP wiring so the installer can call it and users can re-run it after installing a new client).

AlexejPenner (approved):
🦭

AlexejPenner (approved):

.github/workflows/installer.yml
```diff
@@ -0,0 +1,83 @@
...
+        with:
+          persist-credentials: false
+      - name: Run the installer as a user would
+        shell: bash
+        env:
+          KITARU_VERSION: ${{ inputs.version }}
+        # PyPI can lag a few seconds behind a publish; retry the pinned install.
+        run: |
+          for attempt in 1 2 3 4 5; do
+            cat install.sh | bash -s -- --no-login --verbose && exit 0
```
strickvl: [P1] Test the documented non-verbose path

Every primary smoke attempt passes `--verbose`, so it bypasses the default branch in `quiet`. The status-capture bug in `quiet` is a concrete green-while-red example: failures propagate here but become success for `curl ... | bash`. Run the primary attempt without `--verbose`; if it fails, a separate verbose rerun can print diagnostics, but it must not turn the original failure green.

.github/workflows/release.yml
```diff
@@ -160,8 +160,17 @@ jobs:
...
           packages-dir: package-dist
           skip-existing: true
-  publish-deployables:
+  installer-smoke:
+    # The one-liner (kitaru.ai/install) installs from PyPI, so run it on real
+    # runners against the version just published before promoting anything.
     needs: [build, publish-python]
+    permissions:
+      contents: read
+    uses: ./.github/workflows/installer.yml
```
strickvl, line 172: [P1] The release gate tests a different script from the one users fetch

This called workflow pipes the tag's checked-out `install.sh`, while `kitaru.ai/install` serves `install.sh` from `main`. A release tag may be accepted from `develop` or a maintenance branch, so this gate can pass while the public URL is stale, missing, or failing. Keep the local-file matrix for PRs, but make the release delivery check fetch and run `https://kitaru.ai/install` after its `main` source and worker deployment are current.

docs/worker/redirect.mjs
```diff
@@ -63,11 +83,43 @@ function marketingRedirectTarget(pathname) {
...
 
+/**
+ * Serve one file from GitHub main, cached at the edge for five minutes so a
+ * release lands quickly without hammering GitHub.
+ */
+export async function serveInstallFile(request, { url, type }) {
+  if (request.method !== "GET" && request.method !== "HEAD") {
+    return new Response("Method not allowed", { status: 405 });
+  }
+  const upstream = await fetch(url, {
```
strickvl: [P1] Map rejected upstream requests to the promised 502

The fallback only covers a resolved non-OK response. DNS, connection, and timeout failures reject `fetch`, so the worker throws and returns a generic error instead of the documented 502 with the docs pointer. Wrap this fetch in `try/catch`, reuse the same 502 response in the catch path, and add a test whose fetch mock throws.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+die()  { printf '%s✕%s %s\n' "$C_RED" "$C_RESET" "$*" >&2; exit 1; }
+run()  { [ "$KITARU_VERBOSE" = "1" ] && printf '  %s$ %s%s\n' "$C_DIM" "$*" "$C_RESET" >&2; "$@"; }
+quiet() {
+  # Run a command, showing its output only on failure or --verbose.
+  # stdin is /dev/null so a child can never swallow the rest of this script
+  # when it is being piped in from curl.
+  if [ "$KITARU_VERBOSE" = "1" ]; then run "$@" </dev/null; return $?; fi
+  local out; out="$(mktemp)"
+  if "$@" >"$out" 2>&1 </dev/null; then rm -f "$out"; return 0; fi
+  local status=$?
```
strickvl: [P1] Preserve the failed command's status inside `quiet`

After the condition fails, the completed `if` has status 0 because no branch ran, so this captures 0 rather than the child's status. I reproduced a failed `uv tool install` that printed its error, reused Kitaru 0.23.0, printed `Kitaru is installed`, and exited 0. Capture `$?` in an `else` branch before `cat` or `rm`, then test both quiet and verbose modes with a command that exits nonzero.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+
+say ""
+say "${C_ORANGE}◆${C_RESET} ${C_BOLD}Installing Kitaru${C_RESET}"
+say ""
+
+# ---------------------------------------------------------------------------
+# 1. uv
+# ---------------------------------------------------------------------------
+ensure_path() {
+  case ":$PATH:" in *":$1:"*) ;; *) export PATH="$1:$PATH" ;; esac
```
strickvl, line 161: [P1] Put the newly bootstrapped uv ahead of the old one

If an old uv is earlier on `PATH` and `$HOME/.local/bin` already appears later, this helper does nothing. The bootstrap installs a current uv into `$HOME/.local/bin`, but lines 181 and 220 still resolve the old binary, and Bash may also retain its hashed path. I reproduced that ordering with uv 1.0.0 present after bootstrap while the installer still invoked uv 0.4.0. After bootstrap, force the install directory to the front, run `hash -r`, and verify the selected version meets `KITARU_MIN_UV`.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+}
+
+if [ "$KITARU_SKIP_SKILLS" = "1" ]; then
+  note "Skipping skills (--no-skills)"
+else
+  step "Installing Kitaru agent skills ($KITARU_SKILLS_REPO)"
+  if have npx; then
+    # The skills CLI exits 0 even when it installed nothing, so check the
+    # filesystem rather than the exit code.
+    quiet npx -y skills add "$KITARU_SKILLS_REPO" -g -y -a '*' -s '*' || true
```
strickvl: [P1] Do not run a checkout-local `skills` binary

`npx -y skills ...` resolves `./node_modules/.bin/skills` before fetching a package. I reproduced the exact command executing a local test payload from the current directory. Because users are encouraged to run the installer from an agent repository, this lets that checkout supply arbitrary code to a global installer. Run the command from a new empty temporary directory with an explicit package, or use the tarball path for all skill installs.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+    if skills_present; then
```
strickvl: [P1] One old skill is not proof that the upgrade succeeded

The comment already notes that `npx skills` can exit 0 while installing nothing, but `skills_present` accepts one pre-existing `kitaru-investigation/SKILL.md` in any one host directory. A partial or stale install therefore skips the fallback and prints that every agent was updated. Verify the complete required skill set in each intended destination, and take the fallback unless that postcondition holds.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+MCP_BIN="$TOOL_BIN/kitaru-mcp"
+
+if [ "$KITARU_SKIP_MCP" = "1" ]; then
+  note "Skipping MCP registration (--no-mcp)"
+else
+  step "Registering the Kitaru MCP server"
+  registered=0
+  if have claude; then
+    if claude mcp get kitaru >/dev/null 2>&1; then
+      quiet claude mcp remove --scope user kitaru || true
```
strickvl, line 342: [P1] Do not delete a working MCP entry before replacement is safe

Both client paths remove the current `kitaru` entry before attempting the add. If the add fails because of a CLI change, locked config, or write error, a rerun leaves the user with no working entry. Codex `mcp add` already overwrites an existing name, so its remove is unnecessary; for Claude, preserve and restore the old entry on failure or update the config atomically.

install.sh
```diff
@@ -0,0 +1,406 @@
...
+  [ -n "$src" ] || { rm -rf "$tmp"; return 1; }
+  local host
+  for host in "$HOME/.agents" "$HOME/.claude" "$HOME/.codex"; do
+    if [ "$host" != "$HOME/.agents" ] && [ ! -d "$host" ]; then continue; fi
+    mkdir -p "$host/skills"
+    local skill
+    for skill in "$src"/*/; do
+      [ -f "$skill/SKILL.md" ] || continue
+      rm -rf "$host/skills/$(basename "$skill")"
+      cp -R "$skill" "$host/skills/$(basename "$skill")"
```
strickvl: [P2] Propagate failures from the tarball copy

This function is called as an `if`/`elif` condition, which disables Bash `errexit` inside the function. A failed `mkdir`, `rm`, or `cp` can therefore be followed by cleanup and the explicit `return 0`, printing that skills were copied after leaving partial state. Check each destination mutation explicitly and return nonzero after cleanup on the first failure.

strickvl (changes requested):
I focused on the installer runtime and the checks that are supposed to prove it. Requesting changes because the documented default path can turn failed subprocesses into success, and the installer smoke always selects a different branch that hides that defect.

The inline comments are limited to concrete breakage or realistic reliability failures. The highest priority is: fix `quiet`, run the primary smoke without `--verbose`, isolate the `npx` call from the caller's checkout, and make upgrade paths preserve working uv, skills, and MCP state.

I reproduced the quiet-mode false success with a failed `uv tool install` plus an older installed Kitaru, reproduced the old-uv PATH ordering failure, reproduced the stale-skill false success, confirmed the exact `npx` command executes `./node_modules/.bin/skills`, and confirmed a rejected worker fetch escapes instead of returning 502. `bash -n` passes, ShellCheck reports only two informational items, and the 18 existing worker tests pass.

I deliberately did not ask for broad Docker integration coverage or a larger installer abstraction. The comments propose local fixes and focused regression cases.

htahir1:
Thanks both, the `quiet` one especially: a failed `uv tool install` really did end in a green "Kitaru is installed". All of it is addressed in the latest commit. Finding by finding:

**`quiet` status** (strickvl P1, devin): captures the child's status with `|| status=$?`, prints output on failure, returns it. CI now has a step where `--version=99.99.99` must exit nonzero, on the non-verbose path.

**Smoke uses `--verbose`** (strickvl P1): the primary attempt runs without it. On failure a single verbose rerun prints diagnostics inside a log group, and the step still fails.

**Release gate tests the wrong script / blocks publish** (strickvl P1, devin): `installer.yml` has a `source` input. PRs run the checked-out file; release calls it with `source: url`, which does `curl -fsSL https://kitaru.ai/install | bash` pinned to the released version. It no longer gates `publish-deployables`; it fails the workflow as an alert instead.

**Rejected upstream fetch** (strickvl P1): try/catch around the fetch, same 502 body; test with a throwing mock.

**Old uv still resolved** (strickvl P1, devin): after bootstrap the script uses `$HOME/.local/bin/uv` by absolute path everywhere, forces that dir to the front, `hash -r`, and verifies the version against the minimum. Reproduced with uv 0.4.30 in `/usr/local/bin` ahead of `~/.local/bin`: detected, replaced, Kitaru installed with the new one.

**npx from the checkout, `-a '*'`, `skills-release`** (strickvl P1, devin): `npx` is gone. Skills always come from the kitaru-skills tarball (nothing executes from the current directory, no Node needed) into `~/.agents/skills`, plus `~/.claude/skills` and `~/.codex/skills` when that CLI is installed or the directory exists. Only `skills/*/SKILL.md` is copied, so `skills-release` is out.

**Stale skill as proof / partial copy** (strickvl P1 + P2): every `mkdir`/`rm`/`cp` is checked and returns nonzero on failure, and the postcondition is "every skill from the tarball present in every destination". The success line reports the count and destinations.

**MCP remove before add** (strickvl P1): `~/.claude.json` is snapshotted before the replace, `claude mcp get kitaru` confirms the add, and on failure the snapshot is restored with a message saying so. Reproduced with a stub `claude` whose `add` fails: the file comes back byte-identical. Codex's `add` overwrites, so its `remove` is gone.

**`has_tty` false positive** (devin P1): it now actually opens `/dev/tty`. And with Docker up but no terminal, `kitaru login --local` runs with stdin from `/dev/null` so an agent-driven install finishes; only the team-server device flow still needs a terminal.

**Shadowed kitaru** (devin P2): the success line reports `$TOOL_BIN/kitaru --version`, and if a different `kitaru` was reachable on the PATH the user started with, it warns with the path and version.

**Nits**: conflict detection matches `conflict` as well as the `--upgrade` hint; docker permission-denied gets its own hint (docker group); `wget --https-only`; uv installer temp file is trapped. `~/.claude/skills` is created when the `claude` CLI exists, not only when the dir does.

Not done: a note at the top of `/install.md` about GitBook markup, and `docker_running` still treats "not installed" and "daemon down" the same apart from the permission case. Both fine as follow-ups if you want them.

Tested in Docker on ubuntu:24.04, alpine:3.21, bash:3.2 and fedora:41 (full run, six skills, PATH persisted, no `skills-release`, no extra dot-dirs), plus the must-fail regression, the old-uv ordering, the shadowing kitaru, the MCP restore, `$HOME` with a space, and `| dash`. Installer smoke on the five runners is running now.

strickvl (approved):

htahir1:
@strickvl all of your and Devin's findings are addressed in `2d9009bb` (details in the comment above), CI is green on 49/49 including the new must-fail smoke step. Ready for another look.

strickvl:
> @strickvl all of your and Devin's findings are addressed in `2d9009bb` (details in the comment above), CI is green on 49/49 including the new must-fail smoke step. Ready for another look.

yep I looked and tested already and approved.

strickvl (approved):

Issue #907, opened 2026-08-30
Future fuzzing work: coverage-guided and API fuzzing


strickvl:
Most of the fuzzing work considered here has now shipped. This issue now tracks the final decision and activation work for the database-backed semantic suites that are implemented but not yet scheduled.

## Shipped

- #909 established the Hypothesis profiles, importer and MCP properties, and nightly workflow.
- #932 added Schemathesis API fuzzing against a real PostgreSQL-backed server, checking for server errors and response-schema violations.
- #958 added grammar-aware filter fuzzing.
- #1082 added generated importer, evaluator, API-model, redaction, codec, and adapter properties to the bounded PR and deeper nightly selections.
- #1083, #1084, and #1086 strengthened importer normalization, evaluator arithmetic, and LangGraph capture properties. The generated tests found and pinned production defects in Mastra duplicate handling and wide-exponent decimal addition.
- #1087, #1088, and #1089 added PostgreSQL filter semantics, isolated authenticated CRUD sequences, and task-attempt lifecycle sequences.
- #1085 completed generated PydanticAI and OpenAI Agents record/replay coverage. The broader property-testing issue is closed as #910.

The first scheduled nightly after these merges passed: https://github.com/zenml-io/kitaru/actions/runs/35073817741. It exercised the existing `importers`, `evaluators`, `api-models`, `mcp`, `adapters`, `filters`, and `api` jobs.

## Tooling decisions

- **CrossHair:** not adopted. In three matched trials it found no additional branches or counterexamples, while exercising far fewer cases than ordinary Hypothesis.
- **HypoFuzz:** not adopted because its non-commercial license does not cover normal commercial CI use without a separate license. Reconsider only if licensing changes or a focused trial is approved.
- **Atheris / ClusterFuzzLite:** not planned. Their harness and platform cost is not justified while the existing Hypothesis properties still have higher-value semantic gaps.
- **mutmut:** remains out of scope. Mutation testing measures test-suite sensitivity by changing code; it is not an input fuzzer.

## Remaining activation work

The following accepted suites are currently opt-in and are not in `.github/workflows/fuzz-nightly.yml`:

- [ ] Add `fuzz-filters-pg` as a separate PostgreSQL-backed nightly job.
- [ ] Add `fuzz-api-sequences` as a separate PostgreSQL-backed nightly job.
- [ ] Add `fuzz-task-lifecycle` as a separate PostgreSQL-backed nightly job.
[...]

PR #958, opened 2026-09-02
Add grammar-aware filter fuzzing

### LINKED ISSUE #959: Add grammar-aware fuzzing for recursive list filters
Recursive `filter` parameters need coverage beyond schema-driven API fuzzing: exercise valid nested filter JSON through the Agent list boundary and PostgreSQL compilation, while ensuring hostile JSON shapes fail validation rather than crash.

Tracking coverage-guided testing: #907. This work is implemented in #958, which depends on #932.

strickvl:
## Summary

Recursive list filters now get their own nightly property run. It exercises filter JSON through the Agent list boundary and PostgreSQL SQL compilation, so the recursive grammar cannot silently drift outside the schema-based API fuzzer's coverage.

Related: #907
Related: #932

## Reviewer Notes

The valid cases are deliberately limited to the API's accepted depth and condition caps. They cover every boolean node and Agent filter operator without a live database. A separate hostile-shape property permits normal Pydantic validation failures but rejects crashes such as `RecursionError`.

## Reproduction

```bash
just fuzz-filters
```

The nightly profile ran 2,000 examples for each property in 4.16 seconds. `just check` also passed.


AlexejPenner (approved):
🦭

PR #972, opened 2026-09-03
Ephemeral workers

schustmi:
- Add ability to start ephemeral workers for specific jobs from the server, if no live worker is available to claim them
- Ephemeral workers only claim builtin evaluator/importer tasks, which means they only run trusted code

Once we want to enable it:
- Needs extra installed in server image
- Configuration must happen for Pro workspaces


AlexejPenner (approved):
🦭

PR #973, opened 2026-09-03
Add agent-scoped insights

schustmi:
This PR adds agent-scoped insights, records carrying a title, a description, and data of type `text`, `categorical`, or `binned`, created in batches with `client.insights.create(...)` or `POST /api/v1/insights` and exposed through the CLI and MCP.


AlexejPenner (approved):
🦭

PR #975, opened 2026-09-03
SDK Reference Docs workflow: use the lockfile-pinned Wrangler so the sdkdocs site and kitaru.ai worker deploy again

htahir1:
## Summary

The `SDK Reference Docs` workflow (`docs.yml`) deploys two Cloudflare Workers: the generated reference site at sdkdocs.kitaru.ai and the `kitaru-site` redirect worker that owns the kitaru.ai domain (and, since #952, serves `/install` and `/install.md`). It does not publish the docs book; that is GitBook and was never affected.

Every run of that workflow on `main` has failed since 26 Aug at wrangler-action's own Wrangler install: it runs `npm i wrangler@4.95.0` in the working directory, and npm cannot do that at this pnpm workspace root ("Cannot read properties of null (reading 'edgesOut')"). The redirect-worker step comes after it and therefore never ran either, which is why `kitaru.ai/install` was still a 301 after the 0.25.0 release even though `install.sh` was on `main`.

Fix: pin `wrangler` 4.95.0 as a dev dependency of the docs app (already installed by the workflow's `pnpm install --frozen-lockfile`) and put `docs/node_modules/.bin` on PATH before the two deploy steps. The action probes with `npx --no-install wrangler --version`, finds it, and skips its broken install. No ad-hoc package install, so zizmor is happy.

Also drops the temporary lychee exclusion for `kitaru.ai/install`, which answers now.

## Reviewer Notes

One added step in `docs.yml`, gated by the same condition as the deploy steps, plus the docs `package.json`/`pnpm-lock.yaml` entries. Nothing changes in what is deployed. The Aug 26 run (`32977110624`) and the 0.25.0 run (`33784737360`) show the identical npm failure.

Until this is on `main`, pushes to `main` still fail at that step. `kitaru.ai/install` is live today only because I dispatched this workflow on this branch by hand (run `33791186131`), which also published the sdkdocs site from this branch's content.

### Reproduction

Dispatch `SDK Reference Docs` against this branch: "Put the lockfile-pinned Wrangler on PATH", "Deploy production docs", and "Deploy redirect worker" all pass, and `curl -fsSL https://kitaru.ai/install | head -1` returns `#!/usr/bin/env bash`.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -83,6 +83,17 @@ jobs:
         working-directory: docs
       - name: Test redirect worker mapping
         run: node --test docs/worker/redirect.test.mjs
+      - name: Install Wrangler
+        # wrangler-action installs Wrangler with `npm i` in the working directory,
+        # and npm cannot do that at this pnpm workspace root ("Cannot read
+        # properties of null (reading 'edgesOut')"), which has failed every
+        # production deploy since 26 Aug. A global install satisfies the action's
+        # `npx --no-install wrangler --version` probe, so it skips its own install.
+        if: >-
+          steps.sdk_generator.outputs.available == 'true' &&
+          ((github.event_name == 'push' && github.ref == 'refs/heads/main') ||
+          github.event_name == 'workflow_dispatch')
+        run: npm install -g "wrangler@${WRANGLER_VERSION}" && wrangler --version
       - name: Deploy production docs
         # Push deploys only from main; a manual workflow_dispatch deploys whichever
         # branch it runs against, so docs can be shipped to production from develop.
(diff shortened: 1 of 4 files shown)


htahir1:
@strickvl one-step workflow fix, CI green (zizmor included after switching to the lockfile-pinned wrangler). This is what made the `SDK Reference Docs` workflow fail on every main push since 26 Aug, so neither the sdkdocs.kitaru.ai reference site nor the kitaru.ai redirect worker redeployed (the GitBook docs book is separate and was never affected). kitaru.ai/install is live now only because I dispatched the workflow on this branch by hand. Ready for review.

strickvl (approved):
Cherry on top would be somehow to notify us in Discord when this fails? (At the moment I guess this is triggered on push to main, but nobody was notified about it...). I think we have some logic in here (or maybe like the one we do zenml for agent examples) to post to the discord channel?

PR #976, opened 2026-09-03
CI: post to Discord when the SDK Reference Docs deploy or a release fails

htahir1:
## Summary

Follow-up to #975 (merged). That fix made the `SDK Reference Docs` workflow deploy again, but the underlying gap remains: workflows that run on push to `main` have no PR check to turn red, so when the deploy broke on 26 Aug nobody was notified for a week. This adds a Discord ping on failure, as Alex suggested.

## What changed

- **`.github/workflows/notify-failure.yml`** (new, `workflow_call`): one job that posts a short message (what failed, ref, commit, event, actor, run URL) to the SRE Discord channel. Uses plain `curl` against the existing `DISCORD_WEBHOOK_SRE` repo secret rather than a marketplace action, so there is no `@master`-only dependency to carve a zizmor exception for (the zenml repo's `weekly-agent-pipelines-test.yml` needs one for `Ilshidur/action-discord`).
- **`docs.yml`**: final `notify-failure` job, `if: failure() && github.event_name != 'pull_request'`. PR runs already show as a red check, so only the production path (push to main, manual dispatch) pings.
- **`release.yml`**: final `notify-failure` job needing every job in the chain, `if: failure()`. Covers the non-gating `installer-smoke` too, which otherwise fails quietly by design.
- **`.github/zizmor.yml`**: the existing file-scoped `self-repository` exception for `./` reusable-workflow syntax now covers `docs.yml` as well (same actionlint-vs-zizmor conflict release.yml documents).

## What reviewers should focus on

- Whether `failure()` on a job that needs skipped jobs behaves as expected: skipped jobs do not count as failed, so a rehearsal dispatch that skips `publish-python` will not ping unless something actually fails.
- Message content. It deliberately carries no secrets or logs, just the run link.

## Validation

`uv run yamlfix`, `uvx typos`, actionlint (docker `rhysd/actionlint`), `uvx zizmor --config .github/zizmor.yml` all clean. Not exercised end to end: the job only fires on a real failure of a main push or release, and the secret is not available to PR runs.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -129,3 +129,14 @@ jobs:
           packageManager: npm
           wranglerVersion: ${{ env.WRANGLER_VERSION }}
           command: deploy --config wrangler.redirect.toml
+  notify-failure:
+    # PR runs already surface as a red check; only the production deploy path
+    # (push to main, manual dispatch) needs a Discord ping.
+    needs: build-deploy
+    if: failure() && github.event_name != 'pull_request'
+    permissions: {}
+    uses: ./.github/workflows/notify-failure.yml
+    with:
+      title: SDK Reference Docs deploy
+    secrets:
+      DISCORD_WEBHOOK: ${{ secrets.DISCORD_WEBHOOK_SRE }}
--- a/.github/workflows/notify-failure.yml
+++ b/.github/workflows/notify-failure.yml
@@ -1,0 +1,41 @@
+---
+name: Notify failure
+# Reusable job that posts a short message to the SRE Discord channel when a
+# calling workflow fails. Deploy-on-main workflows have no PR to turn red, so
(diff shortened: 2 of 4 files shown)


htahir1:
@strickvl #975 is merged into `develop`, so the next release to `main` carries the Wrangler fix and the SDK Reference Docs deploy (sdkdocs.kitaru.ai plus the kitaru.ai worker) should go green on its own from then on. Until that release, `kitaru.ai/install` stays live from the manual dispatch.

This PR is the "cherry on top" from Alex: a Discord ping to the SRE channel whenever that workflow or a release fails, so a week-long silent breakage cannot repeat. Small and review-only on the YAML; it cannot be exercised from a PR run since the webhook secret is not exposed there.

strickvl (approved):

PR #977, opened 2026-09-03
Add kitaru setup to wire skills and the MCP server into detected coding agents

### LINKED ISSUE #953: Add a kitaru setup command that wires skills and the MCP server into detected coding agents

htahir1:
## Summary

Closes #953. The one-line installer (#952) detects coding agents in bash and runs once, so anyone who installs Claude Code, Cursor, or Codex the day after installing Kitaru gets no skills and no MCP server until they re-run the installer. `kitaru setup` moves that wiring into the CLI so it can be re-run any time, and the installer now calls it instead of carrying its own detection.

## What changed

- **`src/kitaru/cli/setup.py`** (new): the command. Skills come from the `zenml-io/kitaru-skills` tarball, read member by member (no `extractall`, no symlinks, no path escapes) into `~/.agents/skills`, plus `~/.claude/skills` and `~/.codex/skills` when that CLI is on PATH or its home dir exists. MCP registration per client: Claude Code (`claude mcp get` / `remove` / `add --scope user|project`), Codex (`codex mcp add`, which overwrites), Cursor (`.cursor/mcp.json`, in the repo for a project install, in `~` otherwise), Windsurf (`~/.codeium/windsurf/mcp_config.json`). JSON clients are merged, keeping other servers and replacing only `kitaru`. With no client found it prints the snippet to paste.
- **Launch resolution**: inside a project (`sys.prefix` is a `.venv` under a `pyproject.toml`, cwd inside it) the entry is `uv run --directory <project> kitaru-mcp` with Claude Code project scope, matching what the installer already wrote. Otherwise the absolute `kitaru-mcp` next to the interpreter, so a client that does not share the user's PATH still finds it. A missing `kitaru-mcp` is an `invalid_configuration` error with the install hint.
- **Options**: `--mode read-only|standard|destructive` (default `standard`), `--no-skills`, `--no-mcp`; the global `--server` picks the target URL, defaulting to the CLI's selected server, then `http://localhost:8000`.
- **Output**: a rich steps table on a TTY (`src/kitaru/cli/output.py` `_emit_setup`), the usual JSON envelope otherwise, with `steps`, `skills`, and `mcp_snippet`. One failing client is a warning, not a failure; the command only fails when every step failed.
- **`install.sh`**: after installing the package it probes `kitaru schema setup` (offline) and, if present, runs `kitaru setup` with the same `--server`, `--mode`, `--no-skills`, `--no-mcp` it was given. Releases before this command fall through to the previous bash implementation, unchanged, so `--version 0.24.0` keeps working. Closing message gains a "New editor? kitaru setup" line.
- **Docs**: `installation.md` steps 2 and 3 now say the installer runs `kitaru setup` and name Cursor and Windsurf; the by-hand block uses `uv run kitaru setup` instead of `npx skills add`; `agent-native/setup.md` hint explains re-running after a new editor. README step 1 and CHANGELOG updated.
- `tests/cli/test_setup.py`: 12 tests covering skills install and re-run cleanup, tarball path filtering, Claude and Codex command sequences, JSON merge idempotency, project vs user launch resolution, stored-server default, invalid server, download failure. `test_schema.py` adds `setup` to the top-level command set.

## What reviewers should focus on

- **Client detection heuristics.** Cursor and Windsurf are detected by directory existence, not a CLI. A machine with a stale `~/.cursor` gets a config written. I think that is the right trade (harmless file, and the skills dirs already use the same rule) but it is a judgement call.
- **Claude Code scope handling.** `claude mcp get kitaru` succeeding in another scope than the one we register in makes the `remove` fail; we ignore that and let `add` decide. If `add` fails because the name is taken elsewhere, that surfaces as a warning with the CLI's last output line.
- **`resolve_mcp_launch`** requires cwd inside the project for project mode. Running `kitaru setup` from a tool install while standing in a repo stays user mode; the next_actions line tells the user to add Kitaru to the project and re-run.
- The installer probes `schema setup` rather than a version compare, so the handoff is keyed on capability. Tested against 0.25.0 from PyPI (probe exits 2, bash fallback runs) and a locally built 0.26.0.dev0 (handoff runs).

## Validation

```
uv run ruff format --check .   # clean
uv run ruff check .            # 3 findings, all in the untracked examples/end_to_end/ on this machine, not in the diff
uv run ty check                # 11 diagnostics, same untracked directory
uv run pytest tests/cli tests/mcp -q   # 675 passed, 1 skipped, 13 xfailed
uvx typos                      # clean
shellcheck -s bash install.sh  # two pre-existing infos/warnings, none new
```

Docker, Ubuntu 24.04, from the built wheel:
- `uv tool install` then `kitaru setup`: 6 skills downloaded from GitHub into `~/.agents/skills`, no client, snippet printed; with `~/.cursor` present and `--server http://localhost:9000 --mode read-only`: `~/.cursor/mcp.json` written; `kitaru doctor` reports the 6 skills; `--mode yolo` rejected as `invalid_arguments`.
- `uv init` + `uv add` the wheel + fake `claude` on PATH: `uv run kitaru setup` ran `claude mcp add --scope project kitaru -- /root/.local/bin/uv run --directory /work/agent kitaru-mcp --server ... --mode standard`, wrote `.cursor/mcp.json` in the repo, copied skills into `~/.claude/skills` too.
- `install.sh` end to end with `UV_FIND_LINKS` pointing at a 0.26.0.dev0 build: handoff step runs `kitaru setup`, same results; `--quiet --no-mcp` exits 0 with skills installed. Rich table checked on a pty.

Not exercised: real `claude` / `codex` binaries (faked), Windows paths.

## Follow-ups

- Once this ships in a release, the bash skills/MCP sections in `install.sh` can be deleted after the last supported pre-`setup` version drops out of use.
- `kitaru doctor` could list registered MCP clients; today it only reports skills.


htahir1:
@strickvl this is #953, the `kitaru setup` command we talked about when the installer landed. The installer now hands off to it (with a bash fallback for releases that predate the command), so "installed Cursor yesterday, how do I wire it up" becomes one command. Tested in Docker in both project and tool-install modes, details in the body. The client-detection heuristics for Cursor and Windsurf are the part I would most like your eyes on, since importers and adapters aside, you have the most feel for how people actually have these editors set up.

htahir1:
Two small additions from a colleague's fresh install today (ee6f367e):

- The installer printed plain `kitaru login --local` in a shell whose PATH did not yet have `~/.local/bin`, so the next step failed until a new terminal. When `kitaru` does not resolve in the current shell, the closing message now prints `uvx kitaru ...` (uv reuses the tool environment it just installed) and says a new terminal will have plain `kitaru`. Verified in a container with the tool dir deliberately off PATH: `uvx kitaru doctor` runs from the fresh install.
- `kitaru doctor` printed "Kitaru is needs attention." Fixed the headline, and its missing-skills hint now says `kitaru setup` instead of `npx skills add`.

htahir1:
Thanks for the review, all six were real. Addressed in the latest commit:

1. **`--no-mcp` without `kitaru-mcp`**: the launch is now resolved only under `register_mcp`. When it is wanted but `kitaru-mcp` is missing, that is a failed `mcp` step with the install hint and exit 1, and skills still install.
2. **"every client failed" vs "none detected"**: detected and registered are tracked separately. The manual snippet step only appears when nothing was detected; detected-but-all-failed sets exit 1 with a warning to fix and re-run, so `install.sh` surfaces it.
3. **JSON config mode**: rewritten via `tempfile.mkstemp` in the same directory, `os.chmod` to the original mode before `os.replace`, temp removed on any failure. Test asserts a 0600 `mcp.json` stays 0600 and no `.tmp` is left.
4. **Half-updated skills**: each skill is written to a staging dir next to its target and swapped in with a rename, so a failure or Ctrl-C leaves the previous version intact. Outcomes are reported per destination; a write failure in `~/.codex/skills` no longer marks `~/.agents/skills` failed (test covers it). `destinations` is a `Sequence`. Oversized members are skipped and named in the step detail; the archive limit is enforced while streaming rather than after buffering.
5. **Server resolution**: `resolve_target()` like every other command (`--server`, `KITARU_API_URL`, stored server), then `KITARU_LOCAL_URL` as the installer honored, then the default. Test covers all three.
6. **Shadowed Claude entry**: after `claude mcp add`, `claude mcp get kitaru` is read back; if the winning entry does not carry our command, the step fails with a `claude mcp remove` hint instead of reporting done.

Smaller: `--mode` is validated inside `setup()` as well (no more `type: ignore`), `--no-skills --no-mcp` warns "nothing to do", and the docs note that a re-run overwrites local edits under `~/.agents/skills/kitaru-*`. 544 CLI tests pass.

strickvl (approved):

strickvl (commented):
I found **four actionable issues** in PR #977 at `04b0c1d`. All four passed independent validation.

Testing passed: **703 tests, 1 skipped, 13 expected failures**. I also exercised real skill downloads, Cursor config preservation, real Codex registration, Claude scope handling, custom uv environments, JSON output, and terminal output using temporary configurations. GitHub checks are green. Testing used isolated temporary client configurations; no source changes were made.

**I recommend fixing these before merging:**

1. **P1: Setup can report read-only while Claude remains destructive.** [setup.py:538](https://github.com/zenml-io/kitaru/blob/04b0c1d3e7633c130b0337eec75a81167f3864b1/src/kitaru/cli/setup.py#L538) checks only the executable, not its arguments. With the real Claude CLI, an older local entry kept its old server and destructive mode while setup reported success. Verify the complete effective command and arguments, and reject failed readback.

2. **P2: Custom uv environments overwrite global configuration.** [setup.py:268](https://github.com/zenml-io/kitaru/blob/04b0c1d3e7633c130b0337eec75a81167f3864b1/src/kitaru/cli/setup.py#L268) recognizes only environments named `.venv`. A real project using `UV_PROJECT_ENVIRONMENT=…/env` wrote Cursor’s global configuration instead of its project configuration. Honor the configured environment and retain project scope.

3. **P2: Failed Claude replacement removes the working registration.** [setup.py:521](https://github.com/zenml-io/kitaru/blob/04b0c1d3e7633c130b0337eec75a81167f3864b1/src/kitaru/cli/setup.py#L521) removes the existing entry before adding its replacement. If addition fails, the old entry stays deleted. Restore it on failure, as the previous installer did.

4. **P2: Failed skill replacement leaves the skill unavailable.** [setup.py:410](https://github.com/zenml-io/kitaru/blob/04b0c1d3e7633c130b0337eec75a81167f3864b1/src/kitaru/cli/setup.py#L410) moves the old skill aside but never restores it if the final rename fails. Fault injection confirmed the active skill disappeared into a hidden `.old` directory. Roll back the rename on failure.

The full installer flow and Windows behavior were not exercised locally. Add regression tests for these four cases alongside the fixes.

PR #980, opened 2026-09-04
Add import table and post-import evaluators

schustmi:
This PR adds an `import` table and lets an import request name evaluators that score every imported session once the import finishes.

- Pass `evaluators` to `POST /api/v1/imports`, `client.imports.create(...)`, or `kitaru session import --evaluator NAME@VERSION`. One evaluator task per session and evaluator is appended to the import job when the importer completes.
- `POST /api/v1/imports` now returns the import instead of the job. Read `job_id` off it.
- Call `GET /api/v1/imports` and `GET /api/v1/imports/{id}`, `client.imports.get/list/iter`, `kitaru import list` and `kitaru import get`, or the MCP activity read with kind `import`. The import carries `stats` from the parse and `error` when the importer failed.
- Sessions carry `import_id` and can be filtered by it. Task responses carry `import_id` instead of `payload_blob_id` and `agent_id`.
- Deleting an agent deletes its imports. Deleting an importer nulls the import's importer version and cancels a still pending import task at claim. Deleting a payload blob referenced by an import returns 409.


AlexejPenner (approved):
🦭

PR #983, opened 2026-09-04
Adopt whole evaluator invocations under if_missing

### LINKED ISSUE #981: Experiment reruns adopt only one baseline metric per multi-result evaluator under `if-missing`
## Summary

When an experiment run replays a cohort under the default `--baseline-evaluation-mode if-missing`, and an evaluator returns more than one `EvaluationResult` per session, the second and later runs over the same cohort adopt only **one** of that evaluator's baseline evaluation rows per session. The other metrics disappear from the new run's baseline aggregates, and every replay in that run shows `baseline: null` for them. The rows are still in the database; they are just never linked to the new replays.

## Reproduction

1. Register an evaluator that returns two results per session, for example `has_marker` and `single_lookup` from one `evaluate()` call.
2. Record three sessions into a cohort and create an experiment that uses that evaluator plus `kitaru/cost@1`.
3. Start run 1 under a worker and let it settle. The aggregates are complete: `has_marker` baseline `{count: 3, mean: 0.0, pass_rate: 0.0}`, `single_lookup` baseline `{count: 3, mean: 1.0}`, `cost` baseline `{count: 3}`.
4. Start run 2 for the same experiment and cohort with no changes. The aggregates now read `has_marker` baseline `{count: 0, mean: null, pass_rate: null}` and every replay carries `baseline: null` for `has_marker`, while `single_lookup` and `cost` still show `count: 3`.

Which metric survives is whichever row sorts first on `created desc, id desc`; in the observed run the two rows differed by 4 microseconds.

## Cause

`EvaluationRepository.get_latest_evaluation_ids_by_identity` (`src/kitaru/server/adapters/db/repositories/evaluation_repository.py`, around line 325) ranks rows with `row_number()` partitioned by `(session_id, evaluator_version_id, params_hash)` and keeps `rank == 1`, so it returns exactly one evaluation id per identity. One evaluator invocation writes one evaluation row per `EvaluationResult`, so a multi-metric evaluator produces several rows sharing that identity and differing only by metric name.

`replay_pipeline.py` (around lines 116 to 149) then looks up `adoptable.get(identity)`, links the single adopted id to the new replay, and skips scheduling a fresh baseline evaluation for that identity. The aggregates endpoint reads only linked rows, so the unadopted metrics vanish from the run.

Run 1 is unaffected because nothing is adoptable yet and every metric is evaluated fresh.

## Suggested fix

Make the adoption identity include the metric name, or have `get_latest_evaluation_ids_by_identity` return all rows from the latest invocation for an identity (for example by grouping on the invocation that wrote them, or by ranking per `(identity, name)`), and link every one of them to the replay. Add a test with a two-result evaluator and two consecutive runs over the same cohort.

## Evidence

Found during live verification of the Claude Agent SDK adapter on a local stack at `0.25.0+dev`; the adapter is not involved (the bug is in core and reproduces with any multi-result evaluator). Aggregate dumps for run 1 and run 2 are available on request.

schustmi:
Fixes #981. Experiment reruns under `if_missing` now link every evaluation of the latest evaluator invocation to the new replay instead of a single row, so multi-result evaluators keep all their baseline metrics. cc @strickvl

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 - `POST /api/v1/imports` now returns the import instead of the job, with the job in its `job_id`. Import task responses carry `import_id` instead of `payload_blob_id` and `agent_id`.
 
+### Fixed
+
+- `if_missing` baseline scoring now adopts every evaluation an evaluator call produced instead of only one of them, so a rerun's baseline aggregates no longer lose metrics from an evaluator that returns multiple results.
+
 ## [0.25.0] - 2026-09-03
 
 ### Added
--- a/src/kitaru/server/adapters/db/orm/evaluation.py
+++ b/src/kitaru/server/adapters/db/orm/evaluation.py
@@ -171,6 +171,10 @@ class EvaluationORM(UUIDPrimaryKeyMixin, TimestampMixin, Base):
     evaluator_version_id: Mapped[uuid.UUID | None]
     session_id: Mapped[uuid.UUID]
     task_id: Mapped[uuid.UUID | None]
+    # No foreign key, identifies the evaluator call that produced this row
+    # alongside its siblings, and outlives the task the same way
+    # evaluator_version_id does.
+    invocation_id: Mapped[uuid.UUID | None]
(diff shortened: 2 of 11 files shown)


strickvl (approved):
Ok nice yeah I think it's better to use the freshest values.

PR #984, opened 2026-09-04
Add analyzers

### LINKED ISSUE #1002: Add analyzers that run over a set of sessions and write insights
## Problem

Evaluators score one session at a time, so there is no way to run a plugin over a set of sessions and record what it finds about the set. Insights exist as a resource but can only be written by hand through the batch endpoint.

## Expected behavior

A third plugin kind, `analyzer`, that runs once over a set of sessions and writes insights:

- Register analyzers like evaluators, without agent scoping or a provider.
- Name analyzers on an import next to its evaluators. Each analyzer runs as one task in the import job over every session the import created.
- Insights written this way record the analyzer version, the task, and the params that produced them.
- Evaluator and analyzer entrypoints can be `async def`.

Implemented in #984.

schustmi:
This PR adds analyzers, a plugin kind that runs once over a set of sessions and writes insights. Name analyzers on an import next to its evaluators and each one runs as a task in the import job over every session the import created.

- Register one with `kitaru analyzer register NAME --script ... --entrypoint analyzer`, or through `POST /api/v1/analyzers`, `client.analyzers`, and the `kitaru_analyzers_manage` MCP tool. Analyzers are global: no agent scoping, no provider.
- Write the entrypoint as `analyzer(sessions: list[SessionView], **params) -> InsightInput | list[InsightInput]`.
- Pass `analyzers` to `POST /api/v1/imports`, `client.imports.create(...)`, or `kitaru session import --analyzer NAME@VERSION --analyzer-params '{...}'`.
- Insights written this way carry the analyzer version, the task, and the params that produced them. Insight responses expose `analyzer_version_id` and `analyzer_params`.

Closes #1002


strickvl:
One contract question from implementing the model-backed insight analyzer in #982: could analyzer entrypoints support both synchronous and asynchronous callables?

At the moment `call_analyzer()` invokes the entrypoint directly. Our analyzer makes two async OpenAI calls and records async Langfuse observations. Since `run()` already executes inside an event loop, the analyzer cannot safely bridge that with `asyncio.run()`, and a private thread shim would add complexity at every async analyzer.

I think the contract could stay source-compatible by allowing `AnalyzerReturn | Awaitable[AnalyzerReturn]`, making `call_analyzer()` async, and awaiting the returned value only when `inspect.isawaitable(result)`. `run()` would then await `call_analyzer()`.

That also seems consistent with the docs saying that analyzer implementations will often call a model. Would you be open to including async entrypoints in this PR, or is there a reason analyzers should deliberately remain synchronous?

src/kitaru/server/application/services/insight_recording.py
```diff
@@ -0,0 +1,86 @@
...
+        for result in (InsightInput.model_validate(entry) for entry in results)
+    ]
+    if not insights:
+        return
+    # The task's agent can be deleted while it runs. A vanished agent leaves
+    # nothing to record.
+    try:
+        await agent_repository.get(task.agent_id)
+    except NotFoundError:
+        return
```
schustmi: What is this? Completely useless, no?

src/kitaru/task/analyzer.py
```diff
@@ -0,0 +1,154 @@
...
+                    FilterCondition(
+                        field="status",
+                        op=FilterOp.NE,
+                        value=SessionStatus.IN_PROGRESS.value,
+                    ),
+                ]
+            }
+        ),
+        size=1000,
+    )
```
schustmi: Where is the corresponding code for evaluators?

strickvl (approved):

PR #997, opened 2026-09-07
Filter session nodes by type

### LINKED ISSUE #991: No way to filter session nodes by kind; hook and system spans dominate the listing
**What happened?**

A four-turn recorded Claude Agent SDK run produced 36 session nodes, 26 of them `HookEventMessage` and `SystemMessage` spans. Finding the actual `llm_call` and tool nodes means paging past them.

`list_session_nodes` (`src/kitaru/cli/sessions.py`) accepts only `size`, `cursor` and `include_payloads`, so there is no way to filter the listing by node kind.

To be clear, I am not asking for those spans to be dropped — they are useful for forensics. The gap is that there is no way to read a trace without them.

**What did you expect?**

A filter on the node listing, or a default view that collapses hook and system spans with a flag to show them.

**How to reproduce**

1. Record a four-turn Claude Agent SDK session with hooks active.
2. List that session's nodes.
3. Count `HookEventMessage` and `SystemMessage` against `llm_call` and tool nodes.

**Environment**
- Kitaru version: 0.25.0
- kitaru-claude-agent-sdk: 0.1.0rc0 (PR #940 merge commit bcef26cb; not yet on PyPI)
- claude-agent-sdk: 0.2.152
- Python version: 3.12
- OS: macOS 26.6

**Would you like to fix this yourself?**

Yes, if a `--kind` filter on the listing is the shape you want. It touches the params model and the server side, so I would rather agree the interface first than guess.

strickvl:
## What changed?

Session node listings can select LLM, tool, subagent, or span nodes using Kitaru's shared filter-expression syntax through the API, Python and TypeScript SDKs, MCP, and `kitaru session nodes --filter`. Omitting the filter preserves the full trace in ascending index order.

## Why?

Hook and system spans can fill entire pages before any model or tool calls appear. Filtering before pagination lets each page contain the requested work while retaining every recorded node for forensic reads.

Fixes #991.

## Release context

Ship with the next core and TypeScript SDK releases. No database migration or adapter release is needed.

## Reviewer Notes

`SessionNodeListParams` inherits `FilterableListParams`. The application allowlist exposes only `node_type`, with the existing `eq`, `ne`, and `in` operators and boolean composition. Existing filter validation, SQL compilation, SDK encoding, and MCP parameter construction do the work.

Sort remains restricted to `index:asc`. Filtering preserves original node indexes, hidden parent references, and unfiltered cursor compatibility.

## Reproduction

On a session containing both framework spans and LLM/tool calls, compare:

```bash
kitaru session nodes SESSION_ID --size 2
kitaru session nodes SESSION_ID --size 2 --filter '{"field":"node_type","op":"in","value":["llm_call","tool_call"]}'
```

The second command returns up to two matching calls, even when spans precede them. Continue with its returned cursor and the same filter; original indexes and parent references remain intact. The first command still includes spans. `{"field":"node_type","op":"ne","value":"span"}` selects all non-span nodes.

The HTTP reproduction used 36 synthetic nodes: 26 hook/system spans followed by 10 LLM/tool calls. Before the fix, filtering returned HTTP 422. After the fix, the filtered page contains all 10 calls; an unfiltered page still contains the first 20 spans.

## Local checks run

- `just check` and `just mcp-schema-check`.
- Focused PostgreSQL repository, REST, and Python SDK tests; CLI/MCP coverage includes boolean filter serialization and malformed inputs.
- TypeScript SDK: 228 tests passed, typecheck, lint, and generated-schema freshness under Node 22.
- Full Python run: 4,593 passed, 314 skipped, 13 expected failures; its sole failure was an obsolete CLI schema assertion forbidding `--filter`. After correcting that assertion, all 530 CLI tests passed.


src/kitaru/api_models/v1/session_node.py
```diff
@@ -113,6 +113,10 @@ class SessionNodeCreateRequest(RequestModel):
 class SessionNodeListParams(CursorParams):
     """Session node list params."""
 
+    node_type: list[NodeType] = Field(
+        default_factory=list,
+        description="Match any selected type; omitted or empty matches all nodes.",
+    )
```
schustmi: I don't think we need this on this class here? Instead, we should inherit from `FilterableListParams`

schustmi (changes requested):

src/kitaru/api_models/v1/session_node.py
strickvl: Fixed in f27fea58. SessionNodeListParams now inherits FilterableListParams, and node_type is an allowlisted field in the existing expression validator and SQL compiler. CLI, MCP, and both SDKs use the shared filter syntax; sort stays restricted to index:asc. Pagination and hidden parent references are covered by regression tests. Removed the dedicated node_type field and updated the PR example.

schustmi (approved):

PR #1007, opened 2026-09-07
Add provider connections

### LINKED ISSUE #1006: Provider connections for importer credentials
Importers that fetch from a provider API read their credentials from the environment of the worker that claims the task. That ties every credential to one worker process and makes key rotation a per-worker change.

Add a server-side connection resource that holds a provider's non-secret environment values and its secrets, delivered to the task at claim time through the existing secret_env path. An import names a connection, or falls back to the default connection for the importer's provider, or to the worker environment when neither exists. Importer plugins declare a connection_schema so the CLI and dashboard can drive the create form. Import tasks carry a kitaru/provider label so a customer that will not hand over keys can route them to a worker of their own.

schustmi:
Closes #1006


CHANGELOG.md
strickvl, line 1: sort of identical entries in the changelog?

strickvl (changes requested):

strickvl (approved):

PR #988, opened 2026-09-05
Generate insights with selectable analyzers

strickvl:
## What changed?

Imported sessions can produce evidence-backed chart insights through two explicitly selected analyzers: deterministic insights without model credentials, OpenAI-backed insights, or both. Each result includes supporting evidence and a grounded investigation prompt.

This builds on the analyzer framework and provider connections already on `develop`. Frontend onboarding must explicitly request both analyzers; frontend selection, rendering, and the investigation flow remain separate work.

## Why?

The post-import screen should give users a useful finding and a concrete next action. Deterministic code computes numbers, charts, evidence, and copied prompts. The separate OpenAI analyzer uses two bounded model calls to select findings and customize non-numeric copy. Both analyzers retain their own results, even when findings overlap; neither replaces the other's cards.

## Release context

Both analyzers ship in the independent `kitaru-post-import-insights` distribution. The server registers `kitaru/post-import-insights` and `kitaru/openai-post-import-insights` as distinct package entrypoints. Only the latter declares `provider="openai"` and a connection schema requiring `OPENAI_API_KEY`. Workers install the exact package requirement in the task environment; core does not import the implementation.

The package starts at `0.1.0` and pins unreleased core as `kitaru==0.25.0+dev`. Release preparation must replace this development pin with the compatible published core floor, regenerate the plugin lockfile, and align catalog requirements/display versions. Do not publish the package with the development pin. Publish compatible core before the dependent plugin, and make the plugin wheel available before callers select these analyzers. Keep server and worker versions aligned.

OpenAI execution requires an explicit `model` parameter and credentials supplied through an OpenAI connection or an eligible worker environment. A selected OpenAI task without a connection or a credential-equipped worker remains queued; it is not silently skipped or replaced with deterministic analysis. The deterministic analyzer needs neither a model nor an API key. Optional observation uses dedicated `KITARU_INSIGHTS_LANGFUSE_*` credentials.

Related: #984, #982, #1007.

## Reviewer Notes

Analyzer callables receive `list[UUID]`, not preloaded `SessionView` objects. There is no compatibility branch. The runner lists IDs for the import; each analyzer decides which data to fetch. The post-import plugin uses the task-authenticated SDK to fetch, profile, and release one complete trace at a time. Evaluators retain `SessionView`.

[...]

PR #1022, opened 2026-09-08
Make insight providers optional and remove tracing

strickvl:
## What changed?

The deterministic insights analyzer now installs without OpenAI or Langfuse. The OpenAI analyzer selects `kitaru-post-import-insights[openai]`, and insight generation no longer emits Langfuse tracing.

## Why?

Each analyzer should install only the dependencies it needs, including when it runs on a user's infrastructure. This follows the dependency changes discussed with Michael.

Related: #988

## Release context

Include these changes in the initial `kitaru-post-import-insights` release and the corresponding core release. Package versions remain unchanged; the existing development dependency still needs its published core floor during release preparation.

## Reviewer Notes

The deterministic catalog entry keeps its base requirement; only the OpenAI entry requests the extra. Generation receipts, deadlines, and fallback behavior remain intact. The removed `observe` parameter and observer hooks belonged to the unreleased tracing implementation.

The artifact probe now checks both analyzer registrations even though their requirement strings differ. A separate clean environment verifies that the base wheel installs neither provider SDK, then that the OpenAI extra installs OpenAI without Langfuse.

## Reproduction

Run `uv run --no-sync python scripts/smoke_plugin_artifacts.py --package plugins/packages/post-import-insights` after syncing the root development environment. It builds the core and plugin wheels, checks the base and `[openai]` installations separately, loads both analyzer entrypoints, and verifies idempotent default registration. Expect `Plugin artifact smoke passed`; the command fails if the base install includes OpenAI or either install includes Langfuse.

## Local checks run

- Full plugin suite and default-plugin tests: 1,530 passed.
- Root artifact-probe, release-inventory, and default-plugin tests: 108 passed.
- Full `just plugin-artifact-smoke`: passed, including clean base and OpenAI-extra installs.
- Plugin formatting, lint, and type checks: passed.
- All `just check` components passed. After installing missing development extras, the type check passed; the remaining checks ran separately with a writable uv tools directory.
- Direct final diff review found no actionable issues. Live OpenAI calls were not needed or run.


strickvl:
@schustmi This implements the two changes we discussed: OpenAI is now an optional extra selected by the OpenAI analyzer, and the insights plugin's Langfuse dependency and tracing are removed. If you approve it, feel free to merge it in if you want. Thanks!

schustmi (approved):

PR #1035, opened 2026-09-09
Merge changelog entries automatically with the union driver

htahir1:
Two PRs opened the same week both hit a `CHANGELOG.md` conflict against `develop` (#977 and #1004), and both resolutions were the same: keep both sides. This PR makes git do that on its own.

## What changes

- `.gitattributes` gains `CHANGELOG.md merge=union`. On a local merge or rebase, git keeps both sides of a conflicting hunk in the changelog instead of stopping.
- The changelog rule in `CLAUDE.md` explains the driver, its GitHub limitation, and the one thing to check after such a merge.

## Why the changelog conflicts so often

Every PR inserts its entry at the same spot: the first line under a heading in `[Unreleased]`. Git merges by matching surrounding context, so two PRs that both added a line directly below `### Fixed` are two different edits at one location, and git cannot pick an order. Code rarely hits this because PRs touch different functions. The changelog is the one file where every PR edits the same three-line neighborhood.

## Reviewer Notes

The driver only runs in a local git. GitHub's mergeability check does not honor custom merge drivers, so a PR whose only conflict is the changelog will still show as conflicting on GitHub. The fix is unchanged from today, `git merge origin/develop` then push, but the merge now completes without a manual edit. The CLAUDE.md note says exactly that so nobody expects the red badge to disappear.

The trade-off worth knowing: `union` never reports a conflict for this file. If two PRs both edit the same existing line, union keeps both versions silently instead of flagging it. For an append-only file that is the right trade, but it is why the guidance asks for a glance at `[Unreleased]` after a merge. The one duplicate we saw during #977 came from both branches already carrying the same entry once; union would have kept that duplicate too, which is the case the note covers.

I checked that the root `AGENTS.md` carries no changelog rule of its own, so there is nothing to mirror there.

### Reproduction

Simulate two PRs inserting under the same heading, with this branch's `.gitattributes` in place:

```bash
git init -q -b base sim && cd sim
cp /path/to/kitaru/.gitattributes .
printf '## [Unreleased]\n\n### Fixed\n\n- old entry\n' > CHANGELOG.md
git add -A && git commit -qm base
git checkout -qb a && sed -i '4a - entry from PR A' CHANGELOG.md && git commit -qam a
git checkout -q base && git checkout -qb b && sed -i '4a - entry from PR B' CHANGELOG.md && git commit -qam b
git merge a -m merge && cat CHANGELOG.md
```

Without the `.gitattributes` line the merge stops with a conflict. With it, the merge exits 0 and the file lists both new entries above the old one.

Local checks run: typos on the two changed files. No code or tests are affected.

https://claude.ai/code/session_01QAUAys8r9ZkczYYGnnJ5ae

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -5,3 +5,10 @@
 #   pnpm-lock.yaml    -> `pnpm install` (in the relevant subdir)
 uv.lock              merge=binary
 docs/pnpm-lock.yaml  merge=binary
+
+# Every PR inserts its changelog line directly under the same [Unreleased]
+# heading, so two PRs conflict on that spot even though both lines belong.
+# The union driver keeps both sides. It only applies to local merges and
+# rebases; GitHub's conflict check ignores merge drivers, so a PR can still
+# show as conflicting until `git merge origin/develop` is run and pushed.
+CHANGELOG.md         merge=union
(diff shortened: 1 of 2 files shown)


schustmi:
Replaced by #1037

PR #1045, opened 2026-09-09
Skip insight analysis below a session minimum

### LINKED ISSUE #1040: Skip post-import insights below a minimum session count
Skip post-import insight generation when fewer than the required number of sessions are available to the analyzer, with five sessions as the proposed initial minimum. Count eligible sessions from the import, excluding duplicates and failed imports, and return an explicit skipped reason together with the eligible count and threshold so the frontend can explain the outcome; this must remain distinct from analysis failure or a successful analysis with no findings.

Related: zenml-io/kitaru#1028, which covers the zero-session case. The corresponding frontend change should route small imports to Sessions and explain why insights were skipped.

Frontend follow-up: zenml-io/zenml-frontend-monorepo#521.

strickvl:
## What changed?

Built-in post-import insight analyzers skip imports with fewer than five eligible sessions. Set `min_sessions` on an analyzer config to override the minimum for automatic analysis or reruns; custom analyzers default to one.

## Why?

Small imports should not trigger insight generation. Skips carry a reason, eligible count, and threshold so callers can distinguish them from analysis failures and successful analysis with no findings.

Fixes #1040.

## Release context

Core backend change. Frontend follow-up: zenml-io/zenml-frontend-monorepo#521.

## Reviewer Notes

Skipped analysis uses the existing completed task status with `result.status = "skipped"`, without starting a worker. Empty reruns now return a completed job instead of HTTP 409. No database migration or new task status is needed.

The shared check counts eligible sessions stored for the import, including sessions saved before an importer retry. Job settlement uses the existing transition service, preserving its events and analytics. OpenAPI and MCP snapshots are updated within the existing schema budgets.

## Reproduction

1. On an import with one completed session, send `POST /api/v1/imports/{id}/analyze` with `{"analyzers":[{"analyzer":"kitaru/post-import-insights"}]}`.
2. Read `GET /api/v1/jobs/{job_id}/tasks`. The task is completed with `{"status":"skipped","reason":"insufficient_sessions","eligible_sessions":1,"min_sessions":5}`, no error, and no worker attempt.
3. Repeat with `"min_sessions":1` on the analyzer config. The task is queued normally. Five eligible sessions also run without an override; an empty import returns the explicit skip with count zero.

## Local checks run

- `just check`: passed.
- Focused import, analyzer, SDK, CLI, PostgreSQL persistence, and MCP suites: 168 passed; worker regression suite: 9 passed, including skipped tasks across pages.
- Full `just test`: 5,083 passed, 17 skipped, 16 expected failures. Five assertion/schema failures were corrected; their affected suites now pass (76 tests). Two TypeScript failures and three setup errors were caused by missing dependencies; after frozen installation, all 15 TypeScript tests pass.

During the first release smoke check, the deployer should repeat the one-session and five-session cases above and inspect the job/task API results. A skipped task being claimed, an incorrect eligible count, or an all-skipped job remaining pending warrants reverting this change.


src/kitaru/api_models/v1/task.py
```diff
@@ -113,7 +113,10 @@ class TaskResponse(TimestampedResponseModel):
     ended_at: datetime | None = Field(default=None, description="Time execution ended.")
     error: str | None = Field(default=None, description="Error from a failed task.")
     result: JsonValue = Field(
-        description="Task result, diagnostic output on a non-completed task."
+        description=(
+            "Result or diagnostics; skipped analysis: status, reason, "
+            "eligible_sessions, min_sessions."
```
schustmi: Don't like this. This is a generic task response, shouldn't mention any details about something  specific.

src/kitaru/server/adapters/rest/ephemeral_workers.py
```diff
@@ -85,9 +86,14 @@ async def start(self, job_id: uuid.UUID, actor: AuthContext) -> None:
...
         if self._ephemeral_workers is None:
             return
-        tasks, _ = await self._job_service.list_job_tasks(
-            job_id, TaskFilter(), actor=actor
+        tasks = await paginate_all(
+            lambda cursor: self._job_service.list_job_tasks(
+                job_id, TaskFilter(cursor=cursor), actor=actor
+            )
         )
+        tasks = [task for task in tasks if not task.terminal]
```
schustmi: Could we put this as a filter in the above query?

schustmi (changes requested):

src/kitaru/api_models/v1/task.py
strickvl: Fixed in 384916cb7. Restored the generic task-result description and regenerated the schemas.

src/kitaru/server/adapters/rest/ephemeral_workers.py
strickvl: Fixed in 384916cb7. The paginated task query now filters to nonterminal statuses, so completed tasks are excluded by the repository rather than loaded and filtered afterward. The worker tests, including the skipped-task pagination case, pass.

src/kitaru/api_models/v1/replay_config.py
```diff
@@ -80,19 +80,21 @@ class AnalyzerConfig(RequestModel):
     """Analyzer config."""
 
     analyzer: PlainStr = Field(description="Analyzer name.")
+    min_sessions: int | None = Field(
+        default=None,
+        ge=1,
+        description="Session minimum (built-in insights: 5; others: 1).",
```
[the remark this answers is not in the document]
strickvl: Fixed in 384916cb7. Regenerated the TypeScript OpenAPI bindings, including min_sessions. pnpm run generate:check and the core TypeScript typecheck pass.

src/kitaru/server/application/services/import_pipeline.py
```diff
@@ -173,6 +172,10 @@ async def build_analysis_task(
         ),
         on_failure=TaskOnFailure.CONTINUE,
     )
+    task.skip_if_insufficient_sessions(
+        eligible_session_count, analyzer.get_min_sessions(), datetime.now(UTC)
+    )
```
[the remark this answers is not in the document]
strickvl: Confirmed that newly created skip records do not emit ANALYSIS_COMPLETED. Leaving this open pending a decision on telemetry for skips: adding it needs a path that does not advance the job during import fan-out and distinguishes skips from executed analyses. Michael’s requested changes and the generated bindings fix are pushed in 384916cb7.

strickvl:
> Re: comment 5601543715

Review findings processed: fixes pushed in 384916cb7; the telemetry discussion remains open.

schustmi (approved):

PR #1090, opened 2026-09-13
Add Python agent registration helpers

strickvl:
## What changed?

Python users can now create an agent and its initial version through `KitaruClient.register_agent(...)`, with the same API on `KitaruSyncClient`. Both clients also expose `register_agent_version(...)` for adding a version to an existing agent.

Registration remains two create operations rather than an implicit update or rollback. If the agent is created but the version request does not return successfully, `AgentRegistrationError` preserves the agent, exact version request, and idempotency key needed to reconcile an ambiguous response safely.

## Why?

The low-level SDK already supported both requests, but users had to discover and coordinate them through `client.api.agents`. The high-level API now provides a typed result, async/sync parity, pre-dispatch validation, and explicit partial-failure state without hiding the underlying create-only semantics.

The documentation also makes the execution boundary explicit: a `RunSpec` records how a worker should launch the agent, but it does not snapshot source files, dependencies, or the environment.

## Release context

This is an additive core Python SDK API. The changelog fragment, generated-reference allowlist, and hand-written SDK guidance are included; no plugin, frontend, website, or release-order follow-up is required.

## Reviewer Notes

Start with `KitaruClient.register_agent`. The important contract is that both request models and both idempotency arguments are validated before the first POST. The two account-scoped keys must differ, and an omitted version key is generated at this boundary so an `AgentRegistrationError` can expose it.

The failure path deliberately does not delete the new agent or submit a different version request. Retrying the exact request with the retained key is safe within the server's idempotency retention window, including when the server committed the version but the response was lost. The sync methods delegate to the same async implementation.

## Reproduction

1. Start or select a local server with `kitaru login --local`.
2. Run the registration snippet in `docs/book/deploy/sdks.md` with a unique agent name.
3. Run `kitaru agent get <name>` and confirm that the returned agent has version `1`; the snippet prints the same agent and version IDs returned by the server.

The partial-failure cases are deterministic in `tests/client/test_kitaru_client.py`: invalid version input and equal normalized keys make no API call, while a lost version response returns a typed recovery payload and performs no rollback or fresh retry.

## Local checks run

- `uv run --extra server pytest tests/client -q` (394 passed)
- `uv run pytest tests/scripts/test_generate_sdk_docs.py -q` (25 passed, 15 skipped because `fumapy` is unavailable)
- `just check`
- `just docs-lint`

No ongoing operational monitoring is required because this adds client-side convenience methods without changing server behavior or storage. After the next core release, smoke-test one registration against staging and confirm that the agent has exactly one initial version; rollback is a client package rollback and requires no data migration.


src/kitaru/client/client.py
```diff
@@ -65,6 +81,41 @@
...
+    Returns:
+        Normalized key, or None when no key was supplied.
+    """
+    if key is None:
+        return None
+    normalized = key.strip()
+    if (
+        not normalized
+        or len(normalized) > _MAX_IDEMPOTENCY_KEY_LENGTH
+        or not normalized.isprintable()
```
schustmi, line 107: Is this a server-side requirement?

Also, if it is: We should probably add this in the api client even, and add it for all methods that allow specifying one?

schustmi (changes requested):

src/kitaru/client/client.py
strickvl: > Is this a server-side requirement? Also, if it is: We should probably add this in the api client even, and add it for all methods that allow specifying one?

Yes. Addressed in `6e1c642ce`: validation now runs in `KitaruAPIClient.request()` for every explicit idempotency key. `register_agent()` reuses the same validator on both keys before its first request, so an invalid version key cannot leave a new agent without a version.

src/kitaru/client/client.py
```diff
@@ -131,6 +154,124 @@ async def get_agent(self, agent: uuid.UUID | str) -> AgentResponse:
...
+        Returns:
+            Created agent version.
+        """
+        version_request = AgentVersionCreateRequest(
+            display_version=display_version,
+            description=description,
+            run_spec=run_spec,
+            capabilities=capabilities,
+        )
+        idempotency_key = validate_idempotency_key(idempotency_key, "idempotency_key")
```
schustmi: Not necessary here anymore?

schustmi: same above I think?

schustmi (changes requested):

src/kitaru/client/client.py
```diff
@@ -131,6 +154,135 @@ async def get_agent(self, agent: uuid.UUID | str) -> AgentResponse:
...
+        display_version: str | None = None,
+        version_description: str | None = None,
+        capabilities: AgentCapabilities | None = None,
+        agent_idempotency_key: str | None = None,
+        version_idempotency_key: str | None = None,
+    ) -> AgentRegistrationResult:
+        """Create an agent and its initial version.
+
+        The operation sends separate agent and version requests. An
+        ``AgentRegistrationError`` records the created agent, exact version
```
ThibaultDef, line 173: A slight remark: If I understood well, I guess that it is `AgentRegistrationResult` instead of `AgentRegistrationError` in the doc ?

docs/book/deploy/sdks.md
```diff
@@ -23,6 +23,48 @@ async with KitaruAPIClient() as client:
 
 The client reaches everything, including single-session replay creation and blob upload, which the MCP server deliberately leaves out. The concept pages show it in context: [replay a session](../concepts/replay.md), [build a cohort](../concepts/cohorts.md), [start an experiment run](../concepts/experiments.md).
 
+### Register an agent from Python
+
+Use the higher-level `KitaruClient` when you want to create an agent and its initial version together:
+
+```python
+from kitaru.api_models.v1.agent_version import RunSpec, RuntimeCapabilities
```
ThibaultDef, line 31: Really useful ! Thanks

docs/book/deploy/sdks.md
```diff
@@ -23,6 +23,48 @@ async with KitaruAPIClient() as client:
...
+from kitaru.client import AgentRegistrationError, KitaruClient
+
+async with KitaruClient() as client:
+    try:
+        registration = await client.register_agent(
+            "support-agent",
+            RunSpec(
```
ThibaultDef, line 38: One naive thought: would it make sense to allow secrets to be passed through the RunSpec? Does that fit with Kitaru’s current architecture?

strickvl: @schustmi wdyt?

src/kitaru/client/client.py
strickvl: removed the redundant check

src/kitaru/client/client.py
strickvl: Ah yeah I guess that's not so clear in the docs. I'll update...

strickvl: https://github.com/zenml-io/kitaru/commit/53032d0ff see this. Hope that clears it up!

ThibaultDef: Thanks a lot !

docs/book/deploy/sdks.md
schustmi: @strickvl @ThibaultDef The run spec already contains a field called `secret_ids`, which is a list of secret ids. These will be injected into the agent environment as environment variables.

schustmi (approved):

